PR #110, opened 2014-06-01
Update reportcommon.py to use requests

keeleysam:
If requests module is installed, use it instead of urllib2.  Use
munki's client certs to connect to the reporting server if enabled.

### DIFF
--- a/assets/client_installer/reportcommon
+++ b/assets/client_installer/reportcommon
@@ -1,13 +1,11 @@
 #!/usr/bin/python
 # encoding: utf-8
-
+ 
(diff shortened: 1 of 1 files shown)


keeleysam:
I'll probably abandon this with the use of cURL - but will definitely abandon it if gurl.py can be used.

bochoven:
No support for POST requests in gurl.py at the moment I'm afraid. Maybe you could poke Greg to add it?

bochoven:
Resolved with the incorporation of gurl.py in the latest release

Issue #145, opened 2014-07-22
Add support for Bootswatch themes


diwanicki:
Would be nice if there were a way to use the themes and have them persist thru code changes.

http://bootswatch.com

PR #147, opened 2014-08-03
minor edit to gitignore and read me #145

nbalonso:


nbalonso:
Providing the files are named `custom.css` and `custom.js` now with commit c46e649 should accommodate both types of installs.

bochoven:
Thanks, you're on a spree!

PR #189, opened 2014-12-30
Location reporting module

gmarnin:
Provides location information on where the Mac is physical located.

This module uses wifi and the Google map api to determine the approximate latitude, longitude and address of the Mac. Besides tracking down and recovering lost or stolen hardware, we are geographical spread out and this would come in handy to show approximate location.


app/modules/location/scripts/location.sh
```diff
@@ -0,0 +1,59 @@
...
+while read -u 5 MAC SS SSID
+do
+	SSID=`echo $SSID | sed "s/^ *//g" | sed "s/ *$//g" | sed "s/ /%20/g"`
+	MAC=`echo $MAC | sed "s/^ *//g" | sed "s/ *$//g"`
+	SS=`echo $SS | sed "s/^ *//g" | sed "s/ *$//g"`
+	URL+="&wifi=mac:$MAC&ssid:$SSID&ss:$SS"
+done
+IFS=$OLD_IFS
+
+curl -s -A "Mozilla" "$URL" > "$locationfile"
```
bochoven: I think 
```curl -s -A "Mozilla" "$URL" > "$locationfile"```
should read 
```curl -s -A "Mozilla" "$URL" > /tmp/gl_coordinates.txt```

**Comment by bochoven** [comment]
I think activating airport to get the location is a bit too much for my environment. It would also be nice if the script stored a hash of the found ssids, so it would not have to check again when the networks don't change.

**Comment by gmarnin** [comment]
Right, i fixed the `curl -s -A "Mozilla" "$URL" > "$locationfile"` line. 

If activating Wifi is too much please suggest an alternate idea. Code submissions welcomed!

**Comment by bochoven** [comment]
How about 'if wifi is not on, don't check'?

**Comment by keeleysam** [comment]
A better method would be to use CoreLocation, which has been part of OS X since 10.6 and is of course built-in.  You can query it using PyObjC.

**Comment by bochoven** [comment]
@keeleysam You don't happen to have a CoreLocation python script lying around?

**Comment by gmarnin** [comment]
CoreLocation seems like a good idea but to doesn't seem practical to implement. https://twitter.com/gmarnin/status/575977591145820160

**Comment by clburlison** [comment]
I have been looking into something like this as well. Using CoreLocation is possible but kind of messy. [Get-location](https://github.com/lindes/get-location) is a script that can output the location data but it is written in Objective-C and would require us to "set" a compiled binary on the client machines. The next step would be to enable location services, and enable the script to access CoreLocation. University of Utah has two projects on github written in python that could help with the this step but making them a requirement for a module of munkireport seems like a huge draw-back. Maybe we could extract the needed parts but even that seems like a mess. The required projects are: 
https://github.com/univ-of-utah-marriott-library-apple/management_tools/
https://github.com/univ-of-utah-marriott-library-apple/privacy_services_manager/

With those installed two projects installed and the compiled binary from Get-location we can run the following commands to get the location of the machine.

Enables location services & gives the binary script access to the CoreLocation service.

```bash
sudo /usr/local/bin/privacy_services_manager.py enable location
sudo /usr/local/bin/privacy_services_manager.py --admin add location /usr/local/bin/get-location
```

Lastly running the get-location command. Get-location has a few different output formats but for the minimum data needed.

```bash
sudo /usr/local/bin/get-location -f json
```
Which outputs the following. 
``{ "type": "Point", "coordinates": [ 32.86717449, -97.26976564 ] }``

Pro of this approach is that is it backwards compatibly with 10.6 all the way up to 10.10. The big con is the setup for this has quite a few moving parts. 

That's my 2 cents.

bochoven:
There's this short script from @tvsutton you could use instead of get-location:

```python
#!/bin/env python

from subprocess import call
import CoreLocation
manager = CoreLocation.CLLocationManager.alloc().init()
manager.startUpdatingLocation()
coord = manager.location().coordinate()
lat, lon = coord.latitude, coord.longitude
print lat, lon
call(['open', 'http://maps.apple.com/?ll=%s,%s&t=h' % (lat, lon)])
```

clburlison:
Just saw that as well. Then we would just need to give location services access to Python.app. I'm gonna look into pulling out the necessary code from the univ-of-utah-marriott-library-apple repos.

clburlison:
Here is a draft script that should output the necessary data to a plist.
https://gist.github.com/clburlison/6b8164cfd563e351b571

I am running into an issue enabling location services to allow python from the command line.

bochoven:
Script looks nice, but doesn't trigger an authorization request. Maybe you could check with @tvsutton how to do that?

bochoven:
Any progress on this?

bochoven:
Addressed with #379

PR #192, opened 2015-01-19
Additional HTTP Headers/Basic Auth Support for install & submit

morgant:
These updates support additional HTTP headers during both installation & submit (pulled from the Munki configs... either `/Library/Preferences/ManagedInstalls.plist` or `/private/var/root/Library/Preferences/ManagedInstalls.plist`). This is useful if you have additional headers you want or need to send, esp. if you're using [HTTP Basic Auth](https://github.com/munki/munki/wiki/Using-Basic-Authentication) and have Munki & MunkiReport running on the same server (since OS X 10.9 Mavericks and newer only support a single realm for an entire site, unless you're splitting up Munki & MunkiReport to different subdomains or want to significantly modify your Apache configs).

If installing from a server that requires HTTP Basic auth, you may want to run the installer as follows:

    bash-3.2$ sudo /bin/bash -c "$(curl -s -H "Authorization: Basic INSERT_AUTH_HASH_HERE" https://example.com/index.php?/install

### DIFF
--- a/app/modules/bluetooth/scripts/install.sh
+++ b/app/modules/bluetooth/scripts/install.sh
@@ -4,7 +4,7 @@
 CTL="${BASEURL}index.php?/module/bluetooth/"
 
 # Get the scripts in the proper directories
-${CURL} "${CTL}get_script/bluetooth.sh" -o "${MUNKIPATH}preflight.d/bluetooth.sh"
+"${CURL[@]}" "${CTL}get_script/bluetooth.sh" -o "${MUNKIPATH}preflight.d/bluetooth.sh"
 
 # Check exit status of curl
 if [ $? = 0 ]; then
--- a/app/modules/directory_service/scripts/install.sh
+++ b/app/modules/directory_service/scripts/install.sh
@@ -4,7 +4,7 @@
 CTL="${BASEURL}index.php?/module/directory_service/"
 
 # Get the scripts in the proper directories
-${CURL} "${CTL}get_script/directoryservice.sh" -o "${MUNKIPATH}preflight.d/directoryservice.sh"
+"${CURL[@]}" "${CTL}get_script/directoryservice.sh" -o "${MUNKIPATH}preflight.d/directoryservice.sh"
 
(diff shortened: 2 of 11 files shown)


bochoven:
If I understand correctly, you're sending any http headers that are defined in  `ManagedInstalls.plist` to the munkireport server on each request?

morgant:
That is correct.

bochoven:
When someone uses HTTP Basic auth for the munki repo but munkireport is on another server, this change would send the munki headers to any munkireport server.

I would like to have a preference item in /Library/Preferences/Munkireport.plist to signify that adding AdditionalMunkiHeaders is required, so reportcommon.curl only sends the additional headers when necessary.  `{UseMunkiHeaders: True}` ?

morgant:
Makes sense. Similar to your request of @keeleysam on Pull Request #110.

We'd have to add a preference on the server side as well to ensure that it injects the headers when it builds out the install script too. I'll mock up an example for both.

morgant:
@bochoven @keeleysam Actually, while I'm playing with this, should we be keeping these prefs separate or is it worth having a server-side pref that says MunkiReport is running on the same host as Munki? Again, thinking of this as well as PR #110.

bochoven:
I don't think a server side pref is necessary, admins can add/change/remove the setting on the clients (using munki/ARD/etc.)

bochoven:
@morgant Are you still working on this?

morgant:
@bochoven Yes. We're using this in production (it's required for our environment), so I'll review it's state (specifically, the prefs situation).

morgant:
Okay, I've reviewed the code and completed adding & testing a client-side-only `UseMunkiAdditionalHttpHeaders` MunkiReport preference which triggers the sending of Munki's additional HTTP headers. That said, this is from an old branch, so I need to create a new branch from 'wip' and make a clean commit (and probably new pull request). I'll try to have all that wrapped up tomorrow.

morgant:
I have replaced this PR with Pull Request #354.

PR #195, opened 2015-02-01
Changed the german localization a bit, fixed typos

fridomac:
Hello!
I updated the german localization a bit, added the graph buttons.

Bye, Fridolin.


munkireport:
Thanks!

PR #205, opened 2015-03-04
kene101 Power Module - NEW

dingerkingh:
I rebuilt kene101’s power module - this time on the latest branch. I am
currently using it on my server with no issues.

### DIFF
--- a/app/views/widgets/power_battery_condition_widget.php
+++ b/app/views/widgets/power_battery_condition_widget.php
@@ -1,0 +1,51 @@
+		<div class="col-lg-4 col-md-6">
+
+			<div class="panel panel-default">
+
+				<div class="panel-heading" data-container="body" title="Total number of known external displays">
+
+					<h3 class="panel-title"><i class="fa fa-flash"></i> Battery Condition</h3>
+
+				</div>
+
+				<div class="panel-body text-center">
+					<?php
+						$queryobj = new Power_model();
+						$sql = "SELECT COUNT(CASE WHEN condition='Normal' THEN 1 END) AS Normal,
+										COUNT(CASE WHEN condition='Replace Soon' THEN 1 END) AS Soon,
+										COUNT(CASE WHEN condition='Service Battery' THEN 1 END) AS Service,
+										COUNT(CASE WHEN condition='Replace Now' THEN 1 END) AS Now,
+										COUNT(CASE WHEN condition='No Battery' THEN 1 END) AS Missing
+							 			FROM power;";
+						$obj = current($queryobj->query($sql));
+					?>
+				<?if($obj):?>
+					<a href="<?=url('show/listing/power#Replace%20Now')?>" class="btn btn-danger">
+						<span class="bigger-150"> <?=$obj->Now?> </span><br>
+						Now
+					</a>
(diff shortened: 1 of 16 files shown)


app/views/widgets/power_battery_condition_widget.php
```diff
@@ -0,0 +1,51 @@
+		<div class="col-lg-4 col-md-6">
+
+			<div class="panel panel-default">
+
+				<div class="panel-heading" data-container="body" title="Total number of known external displays">
```
bochoven: I think the title should be something else..

app/views/widgets/power_battery_health_widget.php
```diff
@@ -0,0 +1,39 @@
+		<div class="col-lg-4 col-md-6">
+
+			<div class="panel panel-default">
+
+				<div class="panel-heading" data-container="body" title="Total number of known external displays">
```
bochoven: I think the title should be something else..

app/views/widgets/power_battery_condition_widget.php
```diff
@@ -0,0 +1,51 @@
...
+										COUNT(CASE WHEN condition='Service Battery' THEN 1 END) AS Service,
+										COUNT(CASE WHEN condition='Replace Now' THEN 1 END) AS Now,
+										COUNT(CASE WHEN condition='No Battery' THEN 1 END) AS Missing
+							 			FROM power;";
+						$obj = current($queryobj->query($sql));
+					?>
+				<?if($obj):?>
+					<a href="<?=url('show/listing/power#Replace%20Now')?>" class="btn btn-danger">
+						<span class="bigger-150"> <?=$obj->Now?> </span><br>
+						Now
```
bochoven: It would be great if 'Now, Service, Soon and Normal' were localised.

bochoven:
Thanks!

dingerkingh:
A question about localization... Do the actual report and listing pages need to be localized as well or why just the widget page? Is that just what you noticed? If so - I'll do my best to localize but unfortunately it will just be Google translate...

I would also like to see some of the shortcut  <?= replaced with <?php echo to help keep with convention of your current code.. Just haven't gotten to sift through it all to replace them all. Should be able to do that tomorrow.

bochoven:
You should only provide the localization for english and any language you're a native speaker in. Others will do the localisations for their language (hopefully). English is the fallback language.

You should not bother too much about replacing the <?= shortcut, I want to move MR to only use REST API for all views in the near future.

clburlison:
I am getting the following error when I test this commit. 

``Uncaught Exception: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'condition='Normal' THEN 1 END) AS Normal, COUNT(CASE WHEN condition='R' at line 1``

bochoven:
@clburlison which sql backend?

bochoven:
@clburlison I pushed a fix to wip (condition is a MySQL reserved word so should be quoted). Could you check?

clburlison:
@bochoven Tested and the error is fixed. Thanks Arjen!

PR #206, opened 2015-03-05
Add temperature_unit processing to the client tab

rickheil:
While testing the recently added power module (which is awesome, by the way!), I noticed that the temperature was just being pulled as a raw int from the database onto that page, leading to ugly temperature display. I poached the conversions from elsewhere and included logic to not display the temperature at all if no value is present.

Only question / todo - would it be preferable to put the temperature label in the left column to match the ones above it (e.g. mAh), or is the right column ok?


bochoven:
Thanks

PR #208, opened 2015-03-09
Added Safari as the default app to track

nbalonso:
This prevents displaying an Uncaught Exception when nothing has been
specified to be tracked.


bochoven:
Ok, thanks

Issue #178, opened 2014-11-18
Bug? Managed Installs list appears in random order


diwanicki:
On the Managed Software tab (tab_munki) of the Client Detail page, the list of Managed Installs appears to have no particular order... I've noticed this in the past, but never paid much attention to it. Now my list of items has grown large enough for this to be a minor issue when scanning for a particular item on that particular machine.

Would it be possible to order Managed Installs in alpha order by "Display Name"?

bochoven:
Fixed with #211

PR #211, opened 2015-03-09
Make managed installs list sortable

dingerkingh:
The managed installs list table on the managed software tab is now
sortable using dataTables. Enhancement Request #178. Thanks for bearing with me on the pull requests and sorry I am such a gitiot.


bochoven:
Thanks!

PR #222, opened 2015-03-17
Russian translation

little2112:
Added the translation to Russian language file.


bochoven:
There's no need for this file, all localisation takes place in javascript.
I will remove the other php language files to prevent confusion about this.

little2112:
Well, I didn't know that. Thank you!

PR #226, opened 2015-03-24
Profile Reporting Module

dingerkingh:
Created a module for reporting computer profile information. Let me know what you think.

### DIFF
--- a/app/modules/profile/scripts/profile.py
+++ b/app/modules/profile/scripts/profile.py
@@ -1,0 +1,63 @@
+#!/usr/bin/python
+"""
+extracts information about the profiles from system profiler
+"""
+
+import sys
+import os
+import subprocess
+import plistlib
+
+# Skip manual check
+if len(sys.argv) > 1:
+    if sys.argv[1] == 'manualcheck':
+        print 'Manual check: skipping'
+        exit(0)
+
+# Create cache dir if it does not exist
+cachedir = '%s/cache' % os.path.dirname(os.path.realpath(__file__))
+if not os.path.exists(cachedir):
+    os.makedirs(cachedir)
+
+sp = subprocess.Popen(['system_profiler', '-xml', 'SPConfigurationProfileDataType'], stdout=subprocess.PIPE)
+out, err = sp.communicate()
+
+result = ''
+
+# If not profiles are installed - write 0 and exit. 
+if not out:
+	result = ''
+	# Write to disk
+	txtfile = open("%s/profile.txt" % cachedir, "w")
+	txtfile.write(result)
+	txtfile.close()
+	exit(0)
+
+plist = plistlib.readPlistFromString(out)
+
+#loop inside each graphic card
+for profiles in plist[0]['_items']:
+	# add number of profiles installed to result
+	for profile in profiles.get('_items'):
+		for payload in profile.get('_items'):
+			result += 'ProfileUUID = ' + profile.get('spconfigprofile_profile_uuid') + '\n'
+			result += 'ProfileName = ' + profile.get('_name') + '\n'
+			result += 'ProfileRemovalDisallowed = ' + str(profile.get('spconfigprofile_RemovalDisallowed')) + '\n'
+			result += 'PayloadName = ' + payload.get('_name') + '\n'
+			result += 'PayloadDisplayName = ' + payload.get('spconfigprofile_payload_display_name') + '\n'
+			result += 'PayloadData = test\n'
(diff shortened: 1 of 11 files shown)


bochoven:
When running profile.py, I get 

```sh
/usr/local/munki/preflight.d/profile.py
Traceback (most recent call last):
  File "/usr/local/munki/preflight.d/profile.py", line 48, in <module>
    result += 'PayloadDisplayName = ' + payload.get('spconfigprofile_payload_display_name') + '\n'
TypeError: cannot concatenate 'str' and 'NoneType' objects
```

app/modules/profile/scripts/profile.py
```diff
@@ -0,0 +1,63 @@
...
+for profiles in plist[0]['_items']:
+	# add number of profiles installed to result
+	for profile in profiles.get('_items'):
+		for payload in profile.get('_items'):
+			result += 'ProfileUUID = ' + profile.get('spconfigprofile_profile_uuid') + '\n'
+			result += 'ProfileName = ' + profile.get('_name') + '\n'
+			result += 'ProfileRemovalDisallowed = ' + str(profile.get('spconfigprofile_RemovalDisallowed')) + '\n'
+			result += 'PayloadName = ' + payload.get('_name') + '\n'
+			result += 'PayloadDisplayName = ' + payload.get('spconfigprofile_payload_display_name') + '\n'
+			result += 'PayloadData = ' + json.dumps(payload.get('spconfigprofile_payload_data')) + '\n'
```
bochoven: You can add a default when using get() so you can provide a default for values that may not be there:

```python
payload.get('spconfigprofile_payload_display_name', 'No name')
```

See http://www.tutorialspoint.com/python/dictionary_get.htm

app/modules/profile/scripts/profile.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/python
+"""
+extracts information about the profiles from system profiler
+"""
+
+import sys
+import os
+import subprocess
+import plistlib
+import json
```
bochoven: Some people have 10.5 clients, json has been added to python since 10.6

app/modules/profile/scripts/profile.py
```diff
@@ -0,0 +1,63 @@
...
+	result = ''
+	# Write to disk
+	txtfile = open("%s/profile.txt" % cachedir, "w")
+	txtfile.write(result)
+	txtfile.close()
+	exit(0)
+
+plist = plistlib.readPlistFromString(out)
+
+#loop inside each graphic card
```
bochoven: Please remove the references to graphic cards and displays in the comments

bochoven:
Did you test this on different versions of OSX? With and without profiles installed?

bochoven:
The order of the profiles changes every time I run `/usr/local/munki/preflight.d/profile.py`
That way, the hash of `/usr/local/munki/preflight.d/cache/profile.txt` changes on every run, and MR has to send that file on every run. Would it be possible to fix this?

app/modules/profile/scripts/profile.py
gmarnin: FYI, Munki 2 no longer supports 10.5 clients.

app/modules/profile/scripts/profile.py
nbalonso: now I know why this seemed familiar :)

app/modules/profile/scripts/profile.py
bochoven:
```python
try:
    import json
except ImportError:
    print 'No JSON support'
```

dingerkingh: I added a check to see what version of OS X is being run and quit if it was less than 10.7. I believe 10.7 was the first to even support profiles.

dingerkingh:
I have been trying to track down the order of the profiles - however, running:
```system_profiler SPConfigurationProfileDataType```
 does in fact produce a random order. I tried using sorted on the dictionary - but can't seem to get that to work. I'll keep trying...

dingerkingh:
I believe I have got the dictionary sorted and not outputting in a random order.

dingerkingh:
Yes - I have tested on devices with and without profiles. If not profiles installed - the device doesn't show up in the listings. If a device had profiles installed and then are removed - the device deletes itself from the listing as well.

bochoven:
Looks good, I'll take a better look tonight. Thanks!

bochoven:
Thanks!

PR #241, opened 2015-04-22
Update widget labels on battery widget

rickheil:
I didn't find the battery widget labels to be very intuitive - here are my suggestions for new ones. If those don't work, maybe a tooltip with the full text and a short explanation of what each status might mean?


bochoven:
You should change the labels in the en.json file.

rickheil:
Oh, duh, sorry! Closing this one and will submit a PR against that.

PR #242, opened 2015-04-22
Update widget labels

rickheil:
I didn't find the battery widget labels to be very intuitive - here are my suggestions for new ones. If those don't work, maybe a tooltip with the full text and a short explanation of what each status might mean?

(updated with correct file for localization!)


bochoven:
Maybe the battery widget needs a different layout, now the labels cause wrapping

bochoven:
Thanks

PR #256, opened 2015-06-25
German translation for business units

fridomac:
Added german translations for business units and fixed a typo or two


bochoven:
Thanks!

PR #269, opened 2015-07-14
Added additional translations to file

fridomac:
Added a lot of new translations to the file, please check if that´s OK. Also changed the structure to match en.json


bochoven:
Thanks!

PR #274, opened 2015-07-21
Wip - edited output of widgets to be more consistent

poundbangbash:
I adjusted some widgets order of output so the dashboard would have a more consistent look to it.  In the default output of some widgets is error->success left to right so the buttons were red/orange/green.  Others output success->error so they were green/orange/red.  My edits make it consistent error->success so buttons are red->green and make a more unified look.  Below are screencaps of the default vs. edited output.

original

edited


bochoven:
Thanks!

PR #283, opened 2015-07-30
One too many autopkg commands

gmarnin:


bochoven:
I should have re-read my instructions, thanks!

PR #294, opened 2015-08-17
Better detection of Server app

gdarm:
Existing Server app does not indicate a running server, but launchctl knows.


gdarm:
now it's a good pull request. ;)

bochoven:
PR is fine! Did you test this on  10.6 and up?  I don't have Mac OS X server before 10.9

gdarm:
Currently not. But the old one doesn't support 10.6 too...
On Monday I can test this with a 10.6 Srv VM. :)

bochoven:
Not sure what the requirements should be.. I'm not using it due to not having OSX servers

gdarm:
The path of the Server app is different from the script. The services are also different.

PR #318, opened 2015-09-18
Add total client count on dashboard

gmarnin:
Currently, you can only see the total client count on the Clients Report page. Seems obvious that the total client count should be on the Dashboard too.


gmarnin:
Closing, need to see why 11 files changed. Should one be 1 file

PR #319, opened 2015-09-18
Add the total client count on the dashboard

gmarnin:
Currently, you can only see the total client count on the Clients Report page. Seems obvious that the total client count should be on the Dashboard too.

When I do a pull request against the wip branch, git wants to add in 10 extra files. I can't figure out how to sync or remove those extra files or I pulled against master.


bochoven:
Ok I cherry-picked your commit into wip, thanks

PR #323, opened 2015-09-22
Allow for graphs to have reversed colors

n8felton:
Before:

After:


bochoven:
Ehh, why?

n8felton:
Typically a gradient spectrum like this is cool to hot to show "good" to "bad" The more orange/red something is, the worse it is. This helps to show that machines at 7 years old are "worse" than machines at <1 year old.

n8felton:
The "OS breakdown" widget could probably benefit from this as well.

bochoven:
I think you'll need a different color scheme, the current one just rotates over the whole visible spectrum

bochoven:
Thanks!

PR #331, opened 2015-10-08
Adjust ActiveDirectory computer name detection

poundbangbash:
The original directoryservice.sh script assumed the hostname was the AD binding object name for the computer object.  That's not always the case.

I've added code to detect the binding name from the AD config on the machine and use that to look up computer object comments with dscl.

We do not have comments in our computer objects so anyone that can test this change would be appreciated.


bochoven:
I think this only will get tested when merged..

PR #335, opened 2015-10-15
Fixed crashplan_model.php referencinging timemachine

poundbangbash:
Crashplan_module.php was still referencing timemachine's info.  Changed to reference crashplan data.


bochoven:
Already fixed

PR #345, opened 2015-11-05
Dynamic munki directorys

weswhet:
Changed any scripts that were hard coded to use '/Library/Managed Installs' to check ManagedInstalls.plist for the ManagedInstallDir and use it accordingly.
Also cleaned up some of the python script to be more pep8 compliant.
I tested these changes on my instance with a new install and works as planned.


bochoven:
Looks good, but I think you need to add code/defaults to fallback to if the machine does not have munki installed.

bochoven:
Hey Wesley, could you check if your changes also work when creating an Munkireport install package on a machine that has not installed munki?

weswhet:
Yeah, I'll have some changes for when munki isn't installed and I'll try and test this afternoon, got a little side tracked the last couple of days.

weswhet:
@bochoven Let me know what you think :)

app/modules/munkireport/scripts/install.sh
```diff
@@ -1,3 +1,16 @@
...
+munki_install_dir=$(/usr/bin/defaults read /Library/Preferences/ManagedInstalls ManagedInstallDir)
+
+# make sure the munki directory is defined. If not default back to normal
+if [[ "${munki_install_dir}" == "" ]]; then
+  #Default back to normal since munki doesn't appear to be installed
+  setreportpref "munkireport" '/Library/Managed Installs/ManagedInstallReport.plist'
+fi
+
+# configure munkireport to use munki's config
+setreportpref "munkireport" "${munki_install_dir}/ManagedInstallReport.plist"
```
bochoven: This will reset the preference to a false path if munki_install_dir is not found.

PR #354, opened 2015-11-27
Additional HTTP Headers/Basic Auth Support for install & submit

morgant:
An updated version of Pull Request #192:

Submit Munki's additional HTTP headers along with any curl/purl connection, if the 'UseMunkiAdditionalHttpHeaders' MunkiReport pref is set on the client (this is helpful if you're hosting Munki & MunkiReport on the same server while also using [HTTP Basic Auth](https://github.com/munki/munki/wiki/Using-Basic-Authentication)). Also modernizes/improves curl command building in install_script.php & modules' install.sh scripts to correctly handle complex quoted arguments (like HTTP headers).

If installing from a server that requires HTTP Basic auth, you may want to run the installer as follows:

    sudo defaults write /Library/Preferences/MunkiReport UseMunkiAdditionalHttpHeaders -bool YES
    sudo /bin/bash -c "$(curl -s -H "Authorization: Basic INSERT_AUTH_HASH_HERE" https://example.com/index.php?/install

### DIFF
--- a/app/modules/ard/scripts/install.sh
+++ b/app/modules/ard/scripts/install.sh
@@ -7,7 +7,7 @@ PREF_FILE="/Library/Preferences/com.apple.RemoteDesktop.plist"
 CTL="${BASEURL}index.php?/module/${MODULE_NAME}/"
 
 # Get the scripts in the proper directories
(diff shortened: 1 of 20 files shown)


bochoven:
Thanks!

PR #356, opened 2015-12-08
Re-ordering of columns for clients list view

aysiu:
Not sure if you want to merge this in, but this is basically how I approached it on my installation:

1. Add in .js and .css files for the ColReorder plugin for DataTables.
2. Add in two parameters on the clients list view page to call the plugin and allow for a sort-of saved state (local cookie, I believe).


aysiu:
Actually, don't pull it in just yet... it works, but one piece is broken. If you re-order the columns, the formatting for the Uptime and Check-in columns gets messed up if they're not in the original order.

aysiu:
Seems the issue with my approach is that the formatting is coded by column number. I'm going to see if there's a relatively easy way around this...

PR #379, opened 2016-01-26
CoreLocation Module Squashed Commit

clburlison:
This is a complete rewrite of #189 using CoreLocation to locate your Mac. However, we're still using Google's geocoding api for reverse address lookup.

Lots of community members helped make this possible:
@bochoven, @gmarnin, @pudquick, & @arubdesu all directly have code snippets that were used in creating this pull request. So :+1: to them.

This module is limited to 10.8 - 10.11. I wanted to support 10.6/10.7 however they use sqlite databases for Location Services so you'll need to upgrade those systems if you wish to use this module.

On each munki preflight run Location Services will be enabled, and the system Python binary will be given access to Location Services. ~~Due to how Location Services and the locationd daemon interact the first run to enable all the services takes about 50 seconds to run while sequential runs take about 10 seconds. ~~Most of the delay in the first run is a sleep command however that's the only reliable way for all the bits and pieces to sync up.~~ (View current Module Readme for updated info) 

If you're testing and want to reset Location Services to factory settings the following script might help: [reset_location_services.py](https://github.com/clburlison/scripts/blob/master/clburlison_scripts/reset_location_services/reset_location_services.py)

-Cheers, 
Clayton

### DIFF
--- a/app/modules/location/README.md
+++ b/app/modules/location/README.md
@@ -34,12 +34,12 @@ The call out to Google's API can be disabled if you wish. If you disable this ap
 Disable Google API loopups:
 
 ```bash
-sudo defaults write /Library/Preferences/MunkiReport ReportPrefs -dict-add google_api_lookup -bool True
+sudo defaults write /Library/Preferences/MunkiReport ReportPrefs -dict-add google_api_lookup -bool False
 ````
 
 Re-enable Google API lookups (default):
 ```bash
-sudo defaults write /Library/Preferences/MunkiReport ReportPrefs -dict-add google_api_lookup -bool False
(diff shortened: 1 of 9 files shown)


bochoven:
Thanks! Would it be possible to add a preference to skip google address lookup?

clburlison:
I've added the ability to skip the google address lookup. Tested this with a profile and with `defaults`. This is also documented it in the Module Readme.

Added the json api piece for location module and tested that I got output, so surely I did it correct? Feel free to check.

Additional enhancements that would be cool but I don't have the time ATM:
* Remove Googles API call completely and use another piece from CoreLocation (CLGeocoder). Unfortunately, my python isn't there right now. frogor did the hard part that enabled this module to even exist.
* Add a site map that shows all computers.

bochoven:
Thanks!

PR #384, opened 2016-02-11
Added/updated german localization

fridomac:
I added the Location stuff to the german localization....


fridomac:
Fixed it, thanks for reviewing it!

Bye, Frido.

bochoven:
Thanks!

Issue #377, opened 2016-01-13
report_broken_client not respecting passphrases


tuxudo:
The report_broken_client client doesn't respect passphrases used for filtered clients. When attempting to use it to send an event message, result is "a:1:{s:5:"error";s:34:"passphrase is required but missing";}". 

Passphrase is the one in: "defaults read /Library/Preferences/MunkiReport Passphrase"

tuxudo:
This was fixed in PR #595

tuxudo:
Fixed

PR #387, opened 2016-02-19
GSX Integration

tuxudo:
Integrates GSX via filipp's gsxlib to provide warranty lookups. Includes widget, listing, and client tab. Designed for drop in replacement of the warranty module.


bochoven:
Thanks! I would like you to change some stuff before I can merge:

- [ ] move the gsx library to lib
- [ ] move the database queries from the widget to the gsx_model
- [ ] fetch the widget data via a JSON call
- [ ] fetch the client tab data via a JSON call instead of rendering directly in the page
- [ ] could you supply a sample gsx.txt file for testing

tuxudo:
Requested changes have been made. I am unable to fetch the client tab data via a JSON call. That is beyond my skill at this time. Ground work is there, but I am unable to get it to render properly.

bochoven:
Ok, could you lowercase the database fields (which would be necessary if we ever add PostgreSQL as backend)?

tuxudo:
Database fields have been made lowercase and GSX database sample data has been created and added to GSX module directory. Created potential fix for issue #377

bochoven:
Thanks for the sample data, but I was looking for somewhat simpler: as I don't have access to GSX, I would like a sample /cache/gsx.txt file that I could drop in and run postflight on so it could populate the database.

bochoven:
Oh and maybe don't fix unrelated issues in one PR, please file a separate PR for each issue. Otherwise I need to cherrypick commits if I don't agree with a particular change but want to merge the rest.

tuxudo:
The /cache/gsx.txt only contains the date of when the lookup was last run. Similar to how the warranty module lookup works.

bochoven:
Ah, ok! Not very useful then

bochoven:
Thanks!

PR #389, opened 2016-02-22
Last location update

clburlison:
In order to provide better updates for the location module I've split the lookup script to be self contained with a separate open source project located here: https://github.com/clburlison/pinpoint.

By having its own LaunchDaemon we are able to run for longer periods without making modifications to munki's preflight script. The lookup from the `location.py` script was taking as long as all the other preflight scripts combined.

I'm still working on the Wiki page for Munkireport + pinpoint but basically the process is: set pinpoint as an update_for Munkireport. 

I've already deployed these updates internally without issue. The biggest positive to this split is the ability for me to add additional updates to the location script without having to make PR's to Munkireport. The reporting features shouldn't change much here on out.


bochoven:
Thanks

PR #390, opened 2016-02-24
munkiinfo model

erikng:
This should be everything needed for the munkiinfo model and all appropriate data.

If someone could please test the true/false values in the munki_tab it would be greatly appreciated.


bochoven:
Looks ok, but why not collect all Supported ManagedInstalls Keys? https://github.com/munki/munki/wiki/Preferences#supported-managedinstalls-keys

erikng:
We certainly discussed that, but opted to only gather info on values we deemed informational. 

Most items should be trivial to gather.

bochoven:
Yes, only AdditionalHttpHeaders would be a little challenging as it is an array. There may be a better way to store this data though, that will allow for as many keys as you need without altering the database scheme.

Consider:

table munkiinfo:

    id - unique id
    serial_number - non unique string
    key - string name of the key
    value - string value

collecting data from such a structure would require some smartness (get all records for a particular serial_number)

clburlison:
My 2¢ is most of those other keys would just clutter the view. If you'll
like us to pull all the supported keys I'm not opposed to the work. It's
definitely easier to add everything into the database now if that's your
preference.

I would prefer to not list everything in the munki client tab. Many of
those keys would just clutter the view and make finding valuable info
harder. Could be argued that some of those clutter keys are already in the
view.

On Wed, Feb 24, 2016 at 2:54 PM Erik Gomez <notifications@github.com> wrote:

> We certainly discussed that, but opted to only gather info on values we
> deemed informational.
>
> Most items should be trivial to gather.
>
> Sent from my iPhone
>
> > On Feb 24, 2016, at 2:50 PM, Arjen van Bochoven <
> notifications@github.com> wrote:
> >
> > Looks ok, but why not collect all Supported ManagedInstalls Keys?
> https://github.com/munki/munki/wiki/Preferences#supported-managedinstalls-keys
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/munkireport/munkireport-php/pull/390#issuecomment-188449753>
> .
>
-- 
Clayton Burlison
@clburlison
https://clburlison.com

bochoven:
I'm with you Clayton: collect all, show the necessary stuff (maybe hide the rest under a popup panel or something like that)

erikng:
I'm more than happy to reduce the view based on input as well. We can always store the keys for now and figure out a better way to list them.

erikng:
The key is: what's necessary?

tuxudo:
Are you pulling information from /var/root/Library and /Library/?

clburlison:
Yes. We are getting the keys from munkicommon.

On Wed, Feb 24, 2016 at 3:15 PM tuxudo <notifications@github.com> wrote:

> Are you pulling information from /var/root/Library and /Library/?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/munkireport/munkireport-php/pull/390#issuecomment-188455585>
> .
>
-- 
Clayton Burlison
@clburlison
https://clburlison.com

erikng:
We are pulling from munki, using its own logic to grab the data.

erikng:
Please try the latest commit and see if you like this method. If so, I can move whatever elements into it.

erikng:
@bochoven There is some discussion regarding not pulling the HTTP headers for security purposes. If you agree, I propose removing it from the data points and merging, bearing any other requests.

bochoven:
I'm ok with that

bochoven:
Superseded by #400

Issue #371, opened 2015-12-17
Add System Integrity Protection Status


xinput:
Is it possible to add a Report with the status of System Integrity Protection unter the "security" listing?

Maybe by adding the 
csrutil status 

command in one of the preflight scripts?

bochoven:
Could be part of a 'security' module

xinput:
Update:
i found this https://derflounder.wordpress.com/2015/12/18/updated-system-integrity-protection-status-reporting-script/

maybe this could be implemented in some way.

bochoven:
Sure, it would need to be added to a module, I'm not planning on writing one at the moment, but you can give it a shot: https://github.com/munkireport/munkireport-php/wiki/How-to-create-a-module

bochoven:
See the new security module by @gmarnin !

PR #391, opened 2016-02-26
Security Module

gmarnin:
First pass at a security module for issue #371 

Currently this module supports Gatekeeper status and SIP status. Status is reported on the Dashboard and on the Security listing page.


bochoven:
Thanks!

PR #393, opened 2016-02-27
WiFi Module

tuxudo:
Add WiFi Module, includes two widgets, listing, and client tab view.


app/modules/wifi/scripts/uninstall.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+rm -f "${MUNKIPATH}preflight.d/wifi"
```
clburlison: This should also include a:
```bash
rm -f "${CACHEPATH}wifi.txt"
```

bochoven:
The module looks fine here, I have some little remarks:

**wifi_tab.php**
- For the wifi_tab.php could you take a look at location_tab?
- Retrieve the data via a JSON call (I'm trying to remove as much php from the views as possible)
- Split the 4 column table into 2 2-column tables like in location_tab

app/modules/wifi/wifi_model.php
```diff
@@ -0,0 +1,127 @@
+<?php
+class wifi_model extends Model {
+
+	function __construct($serial='')
+	{
+		parent::__construct('id', 'wifi'); //primary key, tablename
+		$this->rs['id'] = 0;
+		$this->rs['serial_number'] = $serial; $this->rt['serial_number'] = 'VARCHAR(255) UNIQUE';
+		$this->rs['agrctlrssi'] = '';
```
bochoven: Please define the columns that are numeric as a number (this will aid with sorting/searching).

    $this->rs['agrctlrssi'] = 0;

tuxudo:
clburlison assisted with finishing up the WiFi module. Also included is JSON enabled GSX client tab

bochoven:
Thanks, good work! Maybe next time you could make a separate PR for GSX changes ;-)

PR #394, opened 2016-02-27
security module - output data on older devices

clburlison:
If less than 10.7 we should output data for Gatekeeper.
If less than 10.11 we should output data for SIP.


bochoven:
thnx!

PR #396, opened 2016-02-27
Adding Printer widget

tuxudo:
Adds printer widget. Similar to the manifests or pending installs widget, for for printer names


bochoven:
Looks fine!

PR #397, opened 2016-02-28
location - Fix mysql error with strict tables

clburlison:
Basically if the key is empty don't attempt to write an empty string.

This only affects MySQL users who have `STRICT_TRANS_TABLES` enabled however as of version 5.7.5 this is the default value.

### DIFF
--- a/app/modules/location/location_model.php
+++ b/app/modules/location/location_model.php
@@ -75,10 +75,6 @@ class Location_model extends Model {
 			{
 				$this->$field = $plist[$search];
 			}
-			else
-			{
-				$this->$field = '';
-			}
 		}
 				    
 		$this->save();
--- a/app/modules/location/location_model.php
+++ b/app/modules/location/location_model.php
@@ -1,82 +1,91 @@
 <?php
 class Location_model extends Model {
 
-	function __construct($serial='')
-	{
-		parent::__construct('id', 'location'); //primary key, tablename
-		$this->rs['id'] = '';
-		$this->rs['serial_number'] = $serial; $this->rt['serial_number'] = 'VARCHAR(255) UNIQUE';
-		$this->rs['address'] = '';
-		$this->rs['altitude'] = 0;
-		$this->rs['currentstatus'] = '';
-		$this->rs['ls_enabled'] = 0;
-		$this->rs['lastlocationrun'] = '';
-		$this->rs['lastrun'] = '';
-		$this->rs['latitude'] = 0.0;
-		$this->rs['latitudeaccuracy'] = 0;
-		$this->rs['longitude'] = 0.0;
-		$this->rs['longitudeaccuracy'] = 0;
-		$this->rs['stalelocation'] = '';
-		
-		// Add indexes
-		$this->idx[] = array('address');
-		$this->idx[] = array('currentstatus');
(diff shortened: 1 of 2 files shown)


bochoven:
Ok, but this would retain the previous value. Not sure this is a real world problem.

clburlison:
Ew. I'm not sure if that would be a problem either. In slack you mentioned.
> You could remove the automatic retrieving of values when instantiating the model

If that sounds like a better solution I'd need guidance on what you mean.

bochoven:
```php

<?php
class Location_model extends Model {
	
	// Defaults array
	private $defaults = array();

	function __construct($serial='')
	{
		parent::__construct('id', 'location'); //primary key, tablename
		$this->rs['id'] = '';
		$this->rs['serial_number'] = $serial; $this->rt['serial_number'] = 'VARCHAR(255) UNIQUE';
		$this->rs['address'] = '';
		$this->rs['altitude'] = 0;
		$this->rs['currentstatus'] = '';
		$this->rs['ls_enabled'] = 0;
		$this->rs['lastlocationrun'] = '';
		$this->rs['lastrun'] = '';
		$this->rs['latitude'] = 0.0;
		$this->rs['latitudeaccuracy'] = 0;
		$this->rs['longitude'] = 0.0;
		$this->rs['longitudeaccuracy'] = 0;
		$this->rs['stalelocation'] = '';
		
		// Store default values
		$this->defaults = $this->rs;
		
		// Add indexes
		$this->idx[] = array('address');
		$this->idx[] = array('currentstatus');

		// Schema version, increment when creating a db migration
		$this->schema_version = 1;
		
		// Create table if it does not exist
		$this->create_table();
		
		if ($serial)
		{
			$this->retrieve_record($serial);
		}
		
		$this->serial = $serial;
		  
	}
	
	// ------------------------------------------------------------------------

	/**
	 * Process data sent by postflight
	 *
	 * @param string data
	 * 
	 **/
	function process($data)
	{		
		require_once(APP_PATH . 'lib/CFPropertyList/CFPropertyList.php');
		$parser = new CFPropertyList();
		$parser->parse($data);
		
		$plist = $parser->toArray();
		
		// Translate location strings to db fields
		$translate = array(
			'Address' => 'address',
			'Altitude' => 'altitude',
			'CurrentStatus' => 'currentstatus',
			'GoogleMap' => 'googlemap',
			'LS_Enabled' => 'ls_enabled',
			'LastLocationRun' => 'lastlocationrun',
			'LastRun' => 'lastrun',
			'Latitude' => 'latitude',
			'LatitudeAccuracy' => 'latitudeaccuracy',
			'Longitude' => 'longitude',
			'LongitudeAccuracy' => 'longitudeaccuracy',
			'StaleLocation' => 'stalelocation');

		// Parse data
		foreach($translate as $search => $field) {
			    
			if (isset($plist[$search]))
			{
				$this->$field = $plist[$search];
			}
			else
			{
				$this->$field = $this->defaults[$field];
			}
		}		
				    
		$this->save();
	}
}
```

clburlison:
Thanks for the sample code. Your solution is much better. I've updated this PR with the requested changed and tested on MySQL 5.7.11 without issue.

bochoven:
Thanks!

PR #398, opened 2016-02-29
dashboard widget documentation additions

gmarnin:
Added wifi_state and printer to the dashboard layout section

### DIFF
--- a/config_default.php
+++ b/config_default.php
@@ -426,6 +426,7 @@
 	|	munki
 	|	power_battery_condition
 	|	power_battery_health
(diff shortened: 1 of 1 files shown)


bochoven:
Thanks, we should add the widgets to one of the 'report' pages so you can at least view them if you don't add them to the dashboard.

clburlison:
The wifi widgets are already added to the Network Report.

gmarnin:
Makes sense to add the printer widget to the configuration report. I'll do a PR

PR #400, opened 2016-02-29
Munkiinfo module take 2

clburlison:
This is the followup PR to #390 

@erikng Is on break this week and I know everyone is pretty interested in getting a new release up.

The biggest changes to this PR is the modification of the database scheme to store the key/value pairs in two columns. This means additional keys can be added in the future just by modifying the `munkiinfo.py` script.


bochoven:
Although this could use some cleaning up (lose the colons in the headings) and the placement of the popup could be a bit better (it drops from my Macbook screen), I'm going to merge this to get more eyes on it. Thanks @erikng and @clburlison

PR #414, opened 2016-03-03
Update location.php

rfitzwater:
made it so map view is not statically set to 500px, but rather fills the viewport.


rfitzwater:
made it so map view is not statically set to 500px, but rather fills the viewport.

bochoven:
Could you please send the Pull Request against the wip branch?

PR #415, opened 2016-03-04
Update location.php

rfitzwater:
made it so map view is not statically set to 500px, but rather fills the viewport.


bochoven:
http://caniuse.com/#search=calc

bochoven:
Thanks

PR #417, opened 2016-03-04
Update fr.json

lolipale:
Last one ... All keys has been translate now
But it seems that some keys are missing (eg. title of printer report, bluetooth report, clients report, copy, print, etc,) See the print screen below
screenshot


bochoven:
Merci beacoup!

clburlison:
@lolipale You're correct, not all of the modules are fully translatable at this time. Some text is still hardcoded.

lolipale:
@clburlison Thanks you for your answer.

PR #419, opened 2016-03-04
preserve curl options when curl-ing submit.preflight

kujeger:
Mainly, this fixes so that submit.preflight is not mysteriously missing from the generated PKG when downloading from a munkireport-php server with an insecure certificate. All other curl-calls preserve --insecure, and therefore work.


bochoven:
Good find! Thanks

kujeger:
cheers, hope this makes it into 2.8 :)

bochoven:
It will

PR #425, opened 2016-03-08
Fix incorrect status on SIP

### LINKED ISSUE #424: SIP Status incorrect
The `SIP Status` in the Security module seems to be incorrect.

here is an example:

SIP is disabled on my mac but on the Security listing it is shown as `Active`

screen shot 2016-03-08 at 12 36 59

clburlison:
he output from csrutil status has a period at the end of enabled/disabled


bochoven:
thnx

Issue #429, opened 2016-03-10
v2.8.2 Bugfixes [Master List]


clburlison:
With the final release of [v2.8.2](https://github.com/munkireport/munkireport-php/releases/tag/v2.8.2) a few minor issues have been found. These issues are not show stoppers for anyone interested in upgrading however they do need fixing.

If attempting to resolve any of these issues please make requests against the current [bugfix branch](https://github.com/munkireport/munkireport-php/tree/bugfix)

- [x] Security Module: SIP not correctly determining Enabled/Disabled status. https://github.com/munkireport/munkireport-php/commit/5e7c9859e648f745497ac094d3ec6d83b469647c

- [x] Location Module: When attempting to build a release package a permissions issue will be shown as it is attempting to modify `/Library/Preferences/pinpoint/location.plist`. 
   ```bash
+ Installing location
mkdir: /Library/Application Support/pinpoint: Permission denied
2016-03-08 14:09:45.466 defaults[5843:82356] Could not write domain /Library/Application Support/pinpoint/location.plist; exiting
2016-03-08 14:09:45.504 defaults[5844:82366] Could not write domain /Library/Application Support/pinpoint/location.plist; exiting
2016-03-08 14:09:45.539 defaults[5845:82372] Could not write domain /Library/Application Support/pinpoint/location.plist; exiting
chmod: /Library/Application Support/pinpoint/location.plist: No such file or directory
```

- [x] Wireless Module: When Internet Sharing is enabled the current output from `sudo /System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport -I ` is not useful. This results in a blank wifi.txt cache file. Client nodes will then show:

   ```bash
Server Error: SQLSTATE[HY000]: General error: 1366 Incorrect integer value: '' for column 'agrctlrssi' at row 1
```
   It's better to handle this error message inside of the wifi script vs at the database level. PR #430

tuxudo:
Created PR for WiFi module bugs

clburlison:
If I'm not mistake that should be the last outstanding bug from 2.8.2. Anything else I've missed?

clburlison:
Closed https://github.com/munkireport/munkireport-php/releases/tag/v2.8.3

PR #433, opened 2016-03-14
Resolve location module attempting to write to a root owned directory…

clburlison:
… on package creation

#429


bochoven:
Creative solution!

clburlison:
It's your solution from the ARD Module. Just took a while before I saw it.

bochoven:
Ha, you're right! I forgot about that..

PR #440, opened 2016-03-17
Fix for WiFi being turned off

tuxudo:
Fixes WiFi module when WiFi is turned off


bochoven:
Is this enough to fix he issue reported? I would think that you need to set the type for empty strings to int on the fields that expect integers.

tuxudo:
Fixed and tested latest patch for WiFi. Writes zeros instead of attempting to write NULLs.

bochoven:
Thanks

PR #441, opened 2016-03-18
Add additional GSX documentation

tuxudo:
Add additional GSX documentation on how to create the certbundle.pem for use with GSX module.


bochoven:
thnx!

PR #443, opened 2016-03-22
Bug fix for 1970 date in GSX data

tuxudo:
Bug found by benjaming on Slack. Thanks Ben for helping test fixes!

Background: Apple localizes the returned date in the GSX API. This causes problems with non-US (mm/dd/yy) dates to appear as 1970 or incorrect for some machines. Problem resolved by specifying local date format in config.php, with default being US date format.


bochoven:
Thanks!

PR #453, opened 2016-03-30
use proper curl command to preserve options...

kujeger:
...the wifi install script does not use the set of options defined, so fails on e.g. self-signed certificates .


bochoven:
Good catch, but could you send a pull request against the bug fix branch?  https://github.com/munkireport/munkireport-php/tree/bugfix

kujeger:
Sure. See #454

PR #454, opened 2016-03-30
use proper curl command to preserve options

kujeger:
the wifi install script does not use the set of options defined, so fails on e.g. self-signed certificates

(Supersedes #453 )


bochoven:
Thanks!

PR #455, opened 2016-03-31
Findmymac Module

clburlison:
Although I'm on all the commits @poundbangbash did the up front work.

This module adds FindMyMac support.

I've tested against MySQL with the strict setting and sqlite. This module is supported on 10.7+ (technically 10.6 doesn't support FMM so it's also supported).

Let's hold off on merging until the following issues are resolved:

> ~~Crap, one more item. As it currently stands I'm writing the FMM output to MR's Cache directory as `findmymac_raw_data.plist`. This isn't being uploaded however it's needed to obtain the email and Owner name. Should I delete this from the file system or just leave it?~~ 

Leaving this file.
 
> ~~@tuxudo: Also, I considered your suggestion from last night and still disagree. The hostname from Apple is a clustered DNS name. I don't believe that's useful. Even for android or other devices. Also, `personid` has no meaning. That isn't used to log into your iCloud account. It's just a UID that Apple has given that account. I still see no value in it. When signing into iCloud you'll be using account your username/password.~~


app/modules/findmymac/findmymac_controller.php
```diff
@@ -0,0 +1,61 @@
...
+   * @author clburlison
+   **/
+  function get_stats()
+  {
+
+      if( ! $this->authorized())
+      {
+          // die('Authenticate first.'); // Todo: return json
+          $out['error'] = 'Not authorized';
+      }
```
bochoven: You should terminate this function if a person is not authorised

app/modules/findmymac/findmymac_controller.php
```diff
@@ -0,0 +1,61 @@
...
+ * @param string $serial serial number
+ **/
+public function get_data($serial_number = '')
+{
+	$obj = new View();
+
+      if( ! $this->authorized())
+      {
+          $obj->view('json', array('msg' => 'Not authorized'));
+      }
```
bochoven: Same as above

bochoven:
Nice module, I'll wait until all issues are resolved!

clburlison:
@bochoven I've fixed the two authorization issues and added the personid & hostname fields. They could serve value in certain circumstances.

poundbangbash:
I've tested the latest commit on a couple clients that have FMM enabled and disabled.  Both report information correctly.

I only have a single authorized user and not sure how to test the authorization adjustments.

-Eric

bochoven:
Thanks!

clburlison:
@poundbangbash, to test the authorization you need to open on of the JSON urls.

IE: `https://munki/index.php?/module/findmymac/get_data/SOME_SERIAL_HERE` or `https://munki/index.php?/module/findmymac/get_stats/`

Now if you try them as a private session not logged in you will see `not authorization` before the above commit you would see see data which is bad.

poundbangbash:
Gotcha.  Works here as well.

PR #456, opened 2016-04-01
Update gsxlib for May 2016 GSX API changes

tuxudo:
May 2016 GSX API changes will break MunkiReport's GSX integration unless updated gsxlib is used.


bochoven:
I assume this is allright as I don't have a way to test this code

PR #476, opened 2016-05-02
Dynamic Munki Support

weswhet:
Currently munkireport is hard coded to munki's default install directory settings. If you change the ManagedInstallDir Preference in munki it breaks munkireport. The proposed changes uses CoreFoundation's CFPreferences to check how munki is configured (if at all) and configure munkireport accordingly.

### DIFF
--- a/app/modules/inventory/scripts/install.sh
+++ b/app/modules/inventory/scripts/install.sh
@@ -5,12 +5,8 @@ DR_CTL="${BASEURL}index.php?/module/inventory/"
 
 # Find out where the munki directory is to set accordingly.
 munki_install_dir=$(/usr/bin/python -c 'import CoreFoundation; print CoreFoundation.CFPreferencesCopyAppValue("ManagedInstallDir", "ManagedInstalls")')
-munki_install_dir_len=$((${#munki_install_dir}-1))
+munki_install_dir=$(echo ${munki_install_dir} | sed 's/\/$//')
 
-if [[ ${munki_install_dir:$munki_install_dir_len:1} == '/' ]]; then
-    #if our custom path has a '/' at the end we should trim it off
-    munki_install_dir=$(echo ${munki_install_dir} | sed 's/.$//')
-fi
 # Get the scripts in the proper directories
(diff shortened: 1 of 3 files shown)


weswhet:
an update from https://github.com/munkireport/munkireport-php/pull/345 to correspond with latest WIP branch since i'm bad at git.

app/modules/inventory/scripts/install.sh
```diff
@@ -3,20 +3,31 @@
...
 
+# Find out where the munki directory is to set accordingly.
+munki_install_dir=$(/usr/bin/python -c 'import CoreFoundation; print CoreFoundation.CFPreferencesCopyAppValue("ManagedInstallDir", "ManagedInstalls")')
+munki_install_dir_len=$((${#munki_install_dir}-1))
+
+if [[ ${munki_install_dir:$munki_install_dir_len:1} == '/' ]]; then
+    #if our custom path has a '/' at the end we should trim it off
+    munki_install_dir=$(echo ${munki_install_dir} | sed 's/.$//')
+fi
 # Get the scripts in the proper directories
```
bochoven: Why not replace the if then with ```sed 's/\/$//'```

app/modules/inventory/scripts/install.sh
```diff
@@ -3,20 +3,31 @@
...
-
-	# Set preference to include this file in the preflight check
-	setreportpref "inventory" "/Library/Managed Installs/ApplicationInventory.plist"
-
+    # Make executable
+    chmod a+x "${MUNKIPATH}postflight.d/inventory_add_plugins.py"
+    # make sure the munki install directory is defined. If not default back to normal
+    if [[ "${munki_install_dir}" == "None" ]]; then
+        # This also intended behavior if munki isn't installed
+        setreportpref "munkireport" '/Library/Managed Installs/ApplicationInventory.plist'
```
bochoven: This should read 

```
setreportpref "inventory" '/Library/Managed Installs/ApplicationInventory.plist'
```

app/modules/inventory/scripts/install.sh
weswhet: you sir are a better 'seder' than I. 😃

bochoven:
Looks good, thanks!

PR #496, opened 2016-06-14
Fix Obsolete machines in GSX helper

tuxudo:
GSX helper didn't properly write data to Obsolete machines' stock db tables


bochoven:
Thanks!

PR #499, opened 2016-06-14
Added WiFi hardware check

tuxudo:
Added WiFi hardware check, because some Macs don't have WiFi, like the Xserve


bochoven:
Thank you.

PR #508, opened 2016-07-15
Fixed disk_info to support Snow Leopard, as the output now mimics new…

Steffan-Ravn:
…er diskutil versions output.


bochoven:
Thanks!

PR #511, opened 2016-07-27
DeployStudio

tuxudo:
Add DeployStudio module, made at the PSU MacAdmins 2016 Hackathon by Tuxudo and n8felton


bochoven:
Looks good, but may need some work to get merged. Couple of things that I noticed:

- [ ] There are no indexes on the database table which will make searching slow
- [ ] Helper should be converted to a class similar to the ones in app/lib/munkireport so it won't pollute the global namespace
- [ ]  for clarity, rename `/usr/local/munki/preflight.d/ds` to `deploystudio`

tuxudo:
I don't know how to make the helper into a class, that's beyond my skill set.

bochoven:
I merged your changes in a new branch: https://github.com/munkireport/munkireport-php/tree/tuxudo-deploystudio and made a couple of changes.

Could you check if that's working as intended?

tuxudo:
It works as intended for me on a new MR test system.

bochoven:
Could you tell me where these values should come from:

```
		$this->rs['dstudio-last-workflow'] = '';
		$this->rs['dstudio-last-workflow-duration'] = '';
		$this->rs['dstudio-last-workflow-execution-date'] = '';
		$this->rs['dstudio-last-workflow-status'] = '';
```

On my system they remain blank.

tuxudo:
Those values are returned from the DeployStudio server. They sometimes are blank, depending it the DS server recorded anything for the serial number it pulled the information for. They should populate for a serial number of a recently imaged Mac that was imaged off of the primary DS server.

bochoven:
You should set indexes on fields you want to search (quickly). Could you add indexes? I'm also wondering if we would need dstudio-host-serial-number as that information is already in the table.

PR #519, opened 2016-08-16
Fix for SSL detection

computeronix:
Sometimes HTTPS is set to "off"

Basically assuming its not secure if it is not equal to "on"

### DIFF
--- a/app/views/auth/login.php
+++ b/app/views/auth/login.php
@@ -10,7 +10,7 @@
 
 								<?php echo conf('sitename'); ?>
 
-								<?php if(empty($_SERVER['HTTPS'])): ?>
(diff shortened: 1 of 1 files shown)


bochoven:
According to the docs http://php.net/manual/en/reserved.variables.server.php: 

```'HTTPS'
Set to a non-empty value if the script was queried through the HTTPS protocol.```

With a note:
```Note that when using ISAPI with IIS, the value will be off if the request was not made through the HTTPS protocol.```

So I think the check should be: $_SERVER['HTTPS']  empty() or not equal to 'off'

computeronix:
Thanks for the doc link!  It is funny because that is originally how I had it then looked at it from the on method

Changed

bochoven:
Thanks!

PR #521, opened 2016-08-16
Clean Up Syntax

gmarnin:
To match all the other dropdown menus


bochoven:
Thanks (I missed that)!

PR #522, opened 2016-08-19
Fixed the link

gmarnin:
Removed a `/` as it was causing a double slash in the url and breaking the link


gmarnin:
Removed the trailing slash from config_default.php on #523 making this PR redundant

Issue #525, opened 2016-08-28
DeployStudio listing syntax error


MichalMMac:
I've just upgrade my testing home munkireport setup to 2.9.0 beta.
When accessing DeployStudio listing I get following error every time:

`Create table 'deploystudio' failed: SQLSTATE[HY000]: General error: 1 near "-": syntax error`

DB type is `sqlite`. `allow_migrations` is set to `TRUE`.
I don't use deploystudio module on clients.

bochoven:
We have to replace the dashes (-) in the column names with underscores (_) to be compatible with sqlite

PR #526, opened 2016-08-28
Fix for issue #525

tuxudo:
Changed database headers to be underscores instead of dashes


bochoven:
Thanks, that was fast!

PR #528, opened 2016-08-30
Kb mouse module

MiqViq:
Initial suggestion for enhanced functionality on reporting USB keyboard(s) and mouse/mice.
Clien script sends to db but model/controller is not working yet.


bochoven:
Superseeded by #601

PR #529, opened 2016-08-30
Bluetooth patch1

### LINKED ISSUE #489: Bluetooth Reporting Broken with "magic" devices
Bug found by @merrymac

So when Apple released the Magic series of bluetooth devices this broke how we are reporting on the new bluetooth devices (old ones are still reporting properly).

The positive is the fix is pretty simple and it's backwards compatible. The negative is I only have three "magic" devices, and they all live on my desk, so this is low priority for me to fix. When I get time I'll make a PR however I can't promise when that will be.

Key points:

- Might as well rewrite it in python because it's gonna be a bunch of parsing.
- `system_profiler -xml SPBluetoothDataType` to get the data set needed. 
- We really only need the `dict['_items']['device_title']` array. Needed keys:
  - `device_batteryPercent` (string value with percentage symbol at the end) Might need to strip the % for ordering in the database
  - `device_isconnected` (attrib_No | attrib_Yes)
  - `device_minorClassOfDevice_string`(Trackpad | Mouse | Keyboard)
  - `device_services` (Apple Wireless Keyboard | Apple Wireless Trackpad | Apple Magic Mouse | Magic Mouse 2 | Magic Trackpad 2 | Magic Keyboard)
- Need a second call to `system_profiler SPBluetoothDataType | grep 'Bluetooth Power' ` because the first call will not have the Bluetooth Power key. Due to a bug with the `-xml` flag. 
- Will likely need to have a database migration for this update
- And we'll need to update the reporting listing (?) I'm not sold on how we want to display, "Is this a Magic device or an older device". **Community feedback requested** 

Additional request from @poundbangbash & @merrymac, send the "model type" of the bluetooth devices up to MR. Currently we are only reporting Trackpad and they would like to know if it's a magical device.

I'm also uploaded sample xml outputs for old and new (magic) devices (the data set has been stripped so only the needed values are show).

[triple_old.txt](https://github.com/munkireport/munkireport-php/files/282646/triple_old.txt)
[triple_new.txt](https://github.com/munkireport/munkireport-php/files/282647/triple_new.txt)

---
References:
http://www.frederikseiffert.de/blueutil/  (not needed but cool) 
https://macadmins.slack.com/archives/munkireport/p1464189574000738

clburlison:
I _think_ this fully resolves the bluetooth issues with "magic" devices. The bluetooth listing is now displaying each device in its own row. 

I'm dropping the old bluetooth table on update because this data changes so frequently it really isn't worth writing the migration.

Also, I'm not super happy with the bluetooth module but it is functional.

Fixes #489


clburlison:
Also since the battery levels change so rapidly we no longer collect data on devices unless they are actually connected to the machine at report time. I believe that to be sane modification.

Issue #537, opened 2016-09-12
Inventory listings client links not working


salomaaj:
Listings > Inventory > search something > choose client = Client with serial number "xxx" does not exist. Seems that the client lookup is using machine friendly name instead of serial number.

PR #538, opened 2016-09-12
Added Serial number column to fix client link

poundbangbash:
Added serial column and adjusted existing links to reference correct columns.
Addresses https://github.com/munkireport/munkireport-php/issues/537


bochoven:
thnx!

PR #539, opened 2016-09-12
Events hover title to "Recently generated events"

poundbangbash:
The hover title for the Events widget on the dashboard was incorrect due to recent changes to only show the last 50 events.  The previous title stated the events listed were from the last hour.


bochoven:
thx

Issue #542, opened 2016-09-13
Divide by 0 error in power_model.php


poundbangbash:
The following errors occured on a machine that did not have a battery installed:

```
[Tue Sep 13 14:27:31.733357 2016] [:error] [pid 18141] [client 111.222.333.444:57131] PHP Warning:  Division by zero in /munkireport-php/app/modules/power/power_model.php on line 100
[Tue Sep 13 14:27:31.733415 2016] [:error] [pid 18141] [client 111.222.333.444:57131] PHP Warning:  Division by zero in /munkireport-php/app/modules/power/power_model.php on line 103
```

Testing `$this->design_capacity` and `$this->max_capacity` if they are > zero would catch this scenario in the code https://github.com/munkireport/munkireport-php/blob/master/app/modules/power/power_model.php#L100-L103

PR #543, opened 2016-09-14
Adjusted for div by 0 error.

poundbangbash:
Addressed divide by 0 errors when no battery is present.

BR https://github.com/munkireport/munkireport-php/issues/542


bochoven:
tnx!

PR #551, opened 2016-09-22
Change installed memory defaults

fimtitzgerald:
Very basic edit/suggestion to display separate counts for 8GB and 16GB by default - given the 15" MBP ships with 16GB by default.


bochoven:
Could you remove site_helper.php from this PR?

fimtitzgerald:
Hmmm - may have made this messier than intended. Am going to close for now and approach again using correct methods at a later time. Apologies!

PR #559, opened 2016-10-14
disk_report: Skip non-mounted volumes

VitosX:
Skips volumes which are not mounted, to prevent reporting wrong free space as zero bytes.


bochoven:
Thanks

PR #561, opened 2016-10-18
Remove `public` from function `isnotzero`

poundbangbash:
When trying to view the Network Location widget the widget is blank.
Looking at the debug log of the browser shows `Uncaught SyntaxError: Unexpected token function` and blames https://github.com/munkireport/munkireport-php/blob/06745ac2c232e9706fa7c9690e160e5cbee380c9/app/views/widgets/network_location_widget.php#L28

Removing the public declaration allows the data to be viewable in the widget again.


clburlison:
Can be closed per https://github.com/munkireport/munkireport-php/commit/2508c137ed243989bdd92c5340c9c70f48226a0a

bochoven:
tx

Issue #291, opened 2015-08-17
Stale Client Cleanup


ghost:
Is there any feature currently to clean up stale clients?

Such as if a client has not reported in x amount of days (configurable via the server config settings) then remove from the inventory and all associated data for that machine.

bochoven:
Someone requested an 'Archive' function, that will retain the data but remove it from all views.

computeronix:
Second this   not sure if anyone has any ideas on this?   Would be useful to lets say clear out any mac computer older than 60 days of last checkin time?

bryanzak:
I would just recommend that it be configurable. Our teachers won't check in over summer. So we'd need 90 or 120 days

Bryan

computeronix:
attempting to implement this will see how it goes in my test environment

computeronix:
Whats the recommended order to delete a machine entry

based on the serial number delete inventory then other tables or does it not really matter on the order?

bryanzak:
60 days wont work in education we often have computers offline all summer. I'd rather see 120 or configurable

gmarnin:
The number of days should be configurable as everyone has different environments

computeronix:
it will be configurable via the config

computeronix:
fyi here is the mysql command for those interested - just need to integrate it now into the PHP code

just replace 90 with however many days old you want.  Anything older than this timeframe will select those machines

SELECT serial_number, timestamp FROM munkireport WHERE UNIX_TIMESTAMP(STR_TO_DATE(timestamp,'%Y-%m-%d %h:%i:%s')) < UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 90 day)) ORDER BY timestamp;

computeronix:
Working on a helper script to wrap this up and will post tonight or tomorrow

computeronix:
this should now be resolved with the latest PR

computeronix:
Correction to query (military time :) )

SELECT serial_number, timestamp FROM munkireport WHERE UNIX_TIMESTAMP(STR_TO_DATE(timestamp,'%Y-%m-%d %H:%i:%s')) < UNIX_TIMESTAMP(DATE_SUB(NOW(), INTERVAL 90 day)) ORDER BY timestamp;

bochoven:
These queries don't work on sqlite

PR #567, opened 2016-10-27
Computer Cleanup Implement

computeronix:
Implements the ability to run a cleanup anytime the URL is accessed (regardless of user authorization)

http://munkireport-url/index.php?/cleanup/machine_cleanup

By default in the config, this is disabled

To enable make clean_days greater than 0 where the value is anything x days and older from the time the script is run will be removed

### DIFF
--- a/app/controllers/cleanup.php
+++ b/app/controllers/cleanup.php
@@ -1,0 +1,115 @@
+<?php
+class cleanup extends Controller
+{
+    public function __construct()
+    {
+
+    }
+    
+    //===============================================================
+    
+    public function index()
+    {
+        echo 'Cleanup';
+    }
+
+    //===============================================================
+    
+    public function machine_cleanup()
+    {
+
+		//Clean days number
+		$clean_days = conf('clean_days');
+		
+		if($clean_days <= 0) {
+			die('Admin:clean: Computer Cleanup Disabled.');
+		}
+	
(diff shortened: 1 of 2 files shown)


computeronix:
Added orphan record support

Whenever the above page is accessed; it will review the entire table list serial numbers (where applicable) and if it does not have an associated record in munkireport table then remove it.

computeronix:
Takes care of https://github.com/munkireport/munkireport-php/issues/291

bochoven:
I can't merge this, I think the issue has to be solved differently:

- I don't want duplicate code that deletes a machine, so
- The current delete_machine routine needs to be moved to a library
- I have been working on a cron-like mechanism in munkireport to do maintenance tasks
- This would need to be able to fit in that mechanism (I know I have to flesh that out)
- I would not use subqueries in each table: just collect all out-of-date serials using one query and then loop over those calling the delete_machine routine.
- I'm not sure what the 'orphan' cleaning is about, I haven't seen that in my env.

-Arjen

computeronix:
No problem; feel free to use the queries as needed to pull what is needed

Orphan occasionally I have had an orphaned record if the device was removed by "other means"

OldManWarren:
Can I implement this in an existing instance of munkireport?  If so where would I start?

PR #571, opened 2016-11-18
Update deploystudio_server sample conf to omit trailing slash

timsutton:
As discussed in Slack, if a trailing slash is used in this conf, the DS module throws an exception, so updating the sample config.php file. Thanks!


bochoven:
tx!

Issue #576, opened 2016-12-11
Servermetrics module broken on 10.11+


tuxudo:
The Servermetrics module is broken on 10.11+ and possibly 10.10. 

Background: OS X Server on 10.11 (and maybe 10.10) no longer make asl logs details their metrics. Everything is now written to a sqlite database at /Library/Server/Metrics/metrics.sqlite

The Server Statistics client tab may also have to be overhauled.

tuxudo:
Submitted fix, module is only broken on 10.11+. I think the Server Statistics client tab is still broken. It just shows a blank page for me on 10.9, 10.10, and 10.11 Macs.

PR #577, opened 2016-12-13
Issue #576 fix - Servermetrics module

tuxudo:
Fix for Servermetrics module on 10.11+ Macs.


app/modules/servermetrics/scripts/servermetrics.py
```diff
@@ -66,6 +66,10 @@ def __main__():
     if getOsVersion() < 10:
         logFormat = 0
 
+    if getOsVersion() < 11:
```
tuxudo: Fixed

app/modules/servermetrics/scripts/servermetrics.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
```
tuxudo: Added null handling.

app/modules/servermetrics/scripts/servermetrics.sh
```diff
@@ -0,0 +1,43 @@
...
+# Crude Code™ by @tuxudo
```
tuxudo: I don't know python :/

bochoven:
Thanks!

PR #578, opened 2016-12-13
Caching Module

tuxudo:
Created caching module and widget


app/modules/caching/caching_model.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Process data sent by postflight
+     *
+     * @param string data
+     *
+     **/
+    public function process($data)
+    {
+        // Delete previous entries
+        $this->deleteWhere('serial_number=?', $this->serial_number);
```
tuxudo: The data is roughly 21 lines per caching hit or event. At its largest, it shouldn't be any larger than what a busy OS X Server uploads for the servermetrics module. The largest I've seen is roughly 512Kb.

tuxudo:
This module needs a bit more work. It's not yet ready to be merged.

tuxudo:
Updated Caching widget. It's as done as I can get it. I'm not able to get the graphing widget working, but it has a JSON API for outputting that data.

bochoven:
Thanks, I don't have a caching server, so it is hard for me to do a proper review. I hope you will be available to fix things as needed :-)

PR #586, opened 2017-01-10
Fixed a typo

gmarnin:
`nAn` should be `An`


clburlison:
Is that a typo? It looks like he is placing a `\n` character or an enter before the string of text.

gmarnin:
Ah, maybe you're correct.

bochoven:
Is this still an issue?

bochoven:
Thanks!

PR #587, opened 2017-01-10
Firmware Escrow Module

gmarnin:
Escrow a random Firmware password. 

This model works similar to the FileVault escrow model in that a [sample script is provided](https://github.com/gmarnin/munkireport-php/blob/firmware-escrow/app/modules/firmware_escrow/script/firmware-set.sh) that will set a random 6 digit alphanumeric password and then send that password to Munkireport. 

Note: This script isn't automatically downloaded by the client. Feel free to customize this script for your own needs and test. 

I know of 2 issues:
- Once a firmware password is set, running `sudo firmwarepasswd -mode` outputs data on 2 lines but currently the model is only picking up the first line. 
- Would like to come up with a method to confirm the password has been successfully escrowed before rebooting and if not save it somewhere locally and retry escrowing later.


app/modules/firmware_escrow/script/firmware-set.sh
```diff
@@ -0,0 +1,106 @@
...
+echo "Submitting Firmware Password Escrow Report"
+
+# add firmware_escrow pref to MunkiReport configuration plist
+/usr/bin/defaults write /Library/Preferences/MunkiReport ReportItems -dict-add firmware_escrow /tmp/firmware_results.txt
+
+# submitting firmware escrow report to Munkireport server
+/usr/local/munki/postflight
+
+# remove firmware_escrow pref from the MunkiReport configuration plist
+/usr/libexec/PlistBuddy -c "Delete :ReportItems:firmware_escrow" /Library/Preferences/MunkiReport.plist
```
bochoven: Plistbuddy is not cfprefsd aware. Please use defaults or python to remove the entry

gmarnin: Defaults can't remove this entry because it's in a array and I don't know enough python to remove it.

assets/locales/en.json
```diff
@@ -1,825 +1,362 @@
 {
-	"about": {
```
bochoven: I can't really see why this file has changed.

tuxudo: I think the whitespace got changed around when I removed the firmware_escrow bits from it

PR #594, opened 2017-02-09
Caching client tab

tuxudo:
Adds client tab for caching module and cleans up listing. 

PS: That fileSize function is super slick


bochoven:
Tx!

PR #595, opened 2017-02-09
Fix for report_broken_client not working when passphrase is set

tuxudo:
Small fix for report_broken_client not reporting when a passphrase is set for MunkiReport on the client machine.


bochoven:
:1:

PR #596, opened 2017-02-09
Issue 554 fix

tuxudo:
Slight addition from inventari's PR 555


bochoven:
Tx, youre on a spree!

PR #598, opened 2017-02-09
More Caching module bugfixes

tuxudo:
Fixed order of dates in caching client tab, client tab load (by making 'is' into '='), made origin on widget blue, clicking on widget loads caching listing, caching now properly shows up in listings drop down


bochoven:
tx!

PR #599, opened 2017-02-09
MunkiReportInfo Module

tuxudo:
A module to report back information on the client's MunkiReport installation


tuxudo:
Use full MunkiReport version and convert to int, sort module array

bochoven:
I didn't think this would be useful, but it actually is quite informative!

PR #608, opened 2017-02-11
Location module dates fixes

tuxudo:
Listing and client tab had bad dates. Fixed with additional checking and moment.js


bochoven:
thnx!

PR #611, opened 2017-02-12
Changed USB exclusion method

tuxudo:
Done based on advice from bochoven


tuxudo:
Did something wrong, way too many files are being changed in that update

PR #615, opened 2017-02-14
USB module fix for Lion

tuxudo:
Fixed by adding vendor variable checks


tuxudo:
Did something wrong again, way too many files updated.

PR #617, opened 2017-02-14
Munkireportinfo version on Lion fix

tuxudo:
Defaults sometimes doesn't write out the version. Corrected by checking for version key and setting if doesn't exist


bochoven:
I don't think I understand what you're doing and why?

tuxudo:
If a client doesn't have the version key in the MunkiReport.plist, MySQL will error because it's trying to set an INT to NULL. This corrects that by setting it to a fake null which is then blanked on the listing and client tab.

bochoven:
But why set it to -9876543 instead of 0?

tuxudo:
A INT that should rarely, if ever, actually be set. That way the UI views don't blank out actual data

bochoven:
Tx!

PR #620, opened 2017-02-14
Update en.json

lolipale:
Hello,

Typing error on munkireport - passprase instead of passphrase
It's seems that we have duplicate key ( "localonlymanifest on munkiinfo)

### DIFF
--- a/assets/locales/en.json
+++ b/assets/locales/en.json
@@ -590,7 +590,7 @@
 	"munkireportinfo": {
 		"baseurl": "MunkiReport URL",
 		"clienttabtitle": "MunkiReport",
-		"passphrase": "Passprase",
+		"passphrase": "Passphrase",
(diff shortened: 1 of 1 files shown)


bochoven:
👍

lolipale:
Helle Arjen,

Could you verify these key : 
"munkiinfo": {
		"softwarerepourl": "Software Repo URL",
		**"localonlymanifest": "Run Type",**
		"duration": "Duration",
		"applecatalog": "Apple Catalog URL",
		"manifest": "Manifest",
		**"localonlymanifest": "Local Only Manifest",**
		"start": "Start"
},

bochoven:
Hmmm, that does not look right!

bochoven:
I pushed a fix, thanks for reporting.

PR #622, opened 2017-02-14
Rewrote and enhanced power module

tuxudo:
Now does all machines, UPSes, separate views for batteries, fixed some bugs over old module


bochoven:
This is a very large PR, it will take me some time to peruse. One thing: I still don't get the -9876543 as default value. I would think that 0 would be sufficient.

tuxudo:
The -9876543 INT makes the views nicer without a lot of zeros all over it. 

Example, not every Mac supports the lidwake pmset variable or has it set. It has two values, 0 or 1. On devices where is isn't set, MySQL would complain about the INT being NULL. 

Setting it to -9876543 makes it into a fake null that MySQL will take, then in the views it is blanked out with JavaScript looking for -9876543 and replacing it with "" making it appear blank or empty in the views.

poundbangbash:
Loaded the power-pmset branch and when attempting to load the power listing I get:

`Migration error: power: SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'sleep_prevented_by'`

rickheil:
mysql should support nulls:

```
mysql> CREATE TABLE `test` (`testint` int(11));
Query OK, 0 rows affected (0.03 sec)
mysql> INSERT INTO test VALUES (0),(1),(NULL);
Query OK, 3 rows affected (0.01 sec)
Records: 3  Duplicates: 0  Warnings: 0
mysql> SELECT * FROM test;
+---------+
| testint |
+---------+
|       0 |
|       1 |
|    NULL |
+---------+
3 rows in set (0.00 sec)
Add Comment
```

If the issue is that the db handlers aren't capable of passing them, then we should fix that. Could be useful for future improvements where data is nonexistent too.

bochoven:
Instead of -9876543, could it also be -1 meaning no value? -9876543 looks like it has a special meaning.

tuxudo:
Some of the data items collected can be -1, making the UI's hide -1 would cause valid data to appear blank. The -9876543 was selected because it likely will not be an actual collected data value.

bochoven:
I really like the data this module is reporting, but the powerinfo.txt is a nightmare to parse correctly. Especially when looking at the future of munkireport, would it be possible to restructure it to something more sensible? ioreg has an xml flag.

tuxudo:
The ioreg data could be formatted to an XML structure, but most of the data is output from pmset and it doesn't have any output formatting. I could change the ioreg bits to putout as XML and adjust the module

bochoven:
I would merge the other bits into the xml and put that through the cfpropertylist parser

tuxudo:
Have the client power.sh do preprocessing to make it into XML for uploading?

bochoven:
Yes!

bochoven:
Has anyone tested this on older OSes?

tuxudo:
I tested it with Lion. The batteries portion works correctly. The power listing, like sleep/schedules/disk sleep time, don't appear to be working. I think Apple changed the output slightly. I'll try to adjust for it

tuxudo:
Made some changes for new module to work with older OS versions

app/modules/power/scripts/install.sh
```diff
@@ -12,7 +12,12 @@ if [ $? = 0 ]; then
 	chmod a+x "${MUNKIPATH}preflight.d/power.sh"
 
 	# Set preference to include this file in the preflight check
-	setreportpref "power" "${CACHEPATH}powerinfo.txt"
+	setreportpref "power" "${CACHEPATH}powerinfo.xml"
```
bochoven: Just a minor nitpick, please use .plist as extension

app/modules/power/README.md
```diff
@@ -1,24 +1,99 @@
...
+* ups_name - VARCHAR(255) - Name of UPS
+* active_profile - VARCHAR(255) - Currently active power profile
+* ups_charging_status - VARCHAR(255) - Charging status of UPS
+* externalconnected - VARCHAR(255) - Is power adapter connected
+* cellvoltage - VARCHAR(255) - Voltage levels of battery cells
+* manufacturer - VARCHAR(255) - Manufacturer of battery
+* batteryserialnumber - VARCHAR(255) - Serial number of battery
+* fullycharged - VARCHAR(255) - Is battery fully charged
+* ischarging - VARCHAR(255) - Is battery charging
+* standbydelay - INT(11) - Minutes until standby
```
bochoven: According to the manage, standby delay is in seconds: "standbydelay specifies the delay, in seconds, before writing the hibernation image to disk and powering off memory for Standby."

tuxudo:
Resolved merge conflicts

PR #623, opened 2017-02-14
Add TouchBar and exclude USB31Bus devices

tuxudo:
Thank you refreshingapathy!


app/modules/usb/usb_model.php
```diff
@@ -126,7 +126,7 @@ function process($plist)
 			}
                         
 			// Check for USB bus devices and simulated USB devices to exclude
-            $excludeusb = array("USB30Bus","USB20Bus","USBBus","USB11Bus","UHCI Root Hub Simulation","EHCI Root Hub Simulation");
+            $excludeusb = array("USB31Bus","USB30Bus","USB20Bus","USBBus","USB11Bus","UHCI Root Hub Simulation","EHCI Root Hub Simulation");
```
bochoven: Wouldn't it be more future proof to exclude all items that match USB\d+Bus?

tuxudo: Possibly, I'm not very good at regex in preg_match and would have to try that

PR #624, opened 2017-02-15
Munkireportinfo Data Type Change

tuxudo:
Changed reportitems datatype to TEXT for larger data entries


bochoven:
I don't think you can put an index on a TEXT column

tuxudo:
You're correct. Forgot to remove the index. It has been removed.

PR #626, opened 2017-02-18
Fans and Temperatures Module

tuxudo:


app/views/listing/displays.php
```diff
@@ -24,7 +24,7 @@
 			<th data-i18n="listing.displays.model" data-colname='displays.model'></th>
 			<th data-i18n="serial" data-colname='displays.display_serial'></th>
 			<th data-i18n="listing.power.manufactured" data-colname='displays.manufactured'></th>
-			<th data-i18n="listing.displays.nativeresolution" data-colname='displays.native'</th>
+			<th data-i18n="listing.displays.nativeresolution" data-colname='displays.native'></th>
```
bochoven: Please put this in a separate PR

bochoven:
I wonder how future proof the 'one column per sensor' is.

tuxudo:
It's fairly future restive. Apple reuses the more standard ones like battery, CPU, and hard drive ones. As new machines are released with new unique sensors, they can be added on as a new column. At the very least, a new machine that has new sensors will still have the basic temperatures listed

PR #630, opened 2017-02-21
appusage Lion Fix

tuxudo:
Failed to get ManagedInstallDir on older OSes


app/modules/appusage/scripts/appusage
```diff
@@ -2,7 +2,7 @@
 
 # Munki database check
 
-MANAGEDINSTALLDIR=$(defaults read /Library/Preferences/ManagedInstalls.plist ManagedInstallDir)
+MANAGEDINSTALLDIR=$(defaults read /Library/Preferences/ManagedInstalls ManagedInstallDir)
```
clburlison: Technically people can set this with a mobileconfig profile. Using

```bash
MANAGEDINSTALLDIR=$(/usr/bin/python -c "from Foundation import CFPreferencesCopyAppValue; print CFPreferencesCopyAppValue('ManagedInstallDir', 'ManagedInstalls')")
```

would be much safer.

PR #643, opened 2017-02-27
Network shares module

tuxudo:
A module that shows mounted network shares


bochoven:
Would you be able to resolve the conflicts?

app/modules/network_shares/locales/en.json
```diff
@@ -0,0 +1,10 @@
+{
+    "clienttab": "Network Shares",
+    "noshares": "No Network Shares",
+    "reporttitle": "Network Shares Report",
+    "name": "Name",
+    "mntfromname": "Path",
+    "fstypename": "Type",
+    "fsmtnonname": "Mount Directory",
+    "automounted": "Automount"
+}
```
bochoven: path, type and name are all available in the main locale file, so I would not duplicate them here

PR #645, opened 2017-03-01
Homebrew module

tuxudo:


app/modules/homebrew/provides.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+return array(
+    'client_tabs' => array(
+        'homebrew-tab' => array('view' => 'homebrew_tab', 'i18n' => 'homebrew.clienttitle', 'badge' => 'homebrew-cnt'),
+    ),
+    'listings' => array(
+        'homebrew' => array('view' => 'hb_listing', 'i18n' => 'homebrew.clienttitle'),
```
bochoven: This view should be hbrew_listing

PR #646, opened 2017-03-02
Hide unused modules from listings menu in head.php

nathanperkins:
Adds a conf item called conf('show_unused_modules') with default setting of TRUE.

When set to FALSE, the Listings menu will show only modules that are active in conf['modules'].

screen shot 2017-03-01 at 5 05 18 pm

screen shot 2017-03-01 at 5 04 47 pm


nathanperkins:
As a sidenote, @poundbangbash suggested hiding unused modules based on whether there are items in the database table associated with the module.

If that is preferred, I could try implementing that instead.

clburlison:
I love this feature but am not in love with the implementation. 

* I believe this feature would be much more clear if you changed the variable to `hide_nonactive_modules`. 
* 2nd, the default value in `config_default.php` should be `false` to match current MR implementation 
* Lastly the wording after the variable change would be 
   >  When true, modules that are not in conf['modules'] will not be shown in the 'Listings' drop down menu.

Obviously, these are all my own opinions. If you disagree or @bochoven wants changes go with his opinion

Overall 👍

rickheil:
Oh my lord yes this is excellent. I think looking to conf['modules'] is appropriate - there may be modules that I tested with a few machines and the database has some tables and data, that I do not end up deploying widely.

Issue #653, opened 2017-03-14
FR: Wifi add SNR


mikaellofgren:
Hi!
In Wifi report, add Signal to Noise (SNR) like Meraki has
that calculates the value between RSSI and Noise to get the SNR value.

Also great as a widget aswell.

https://documentation.meraki.com/MR/WiFi_Basics_and_Best_Practices/Wireless_fundamentals%3A_Signal-to-Noise_Ratio_(SNR)__and_wireless_signal_strength

mikaellofgren:
Thanks, that was quick. =)

PR #654, opened 2017-03-14
Cleaned up WiFi views

tuxudo:
Added signal noise ratio for issue #653 and localized data


bochoven:
That's a quick response! 😊

Issue #662, opened 2017-03-16
Disk VolumeType Widget


gmarnin:
Feature request for a widget to show the total number of fusion, hdd, ssd, bootcamp and raid drives.

MR determines the VolumeType [here](https://github.com/munkireport/munkireport-php/blob/06745ac2c232e9706fa7c9690e160e5cbee380c9/app/modules/disk_report/disk_report_model.php#L131)

PR #663, opened 2017-03-16
Add disk type widget

tuxudo:
From feature request #662


app/modules/disk_report/disk_report_controller.php
```diff
@@ -65,6 +65,27 @@ public function get_filevault_stats($mount_point = '/')
...
+    {
+        $obj = new View();
+
+        if (! $this->authorized()) {
+            $obj->view('json', array('msg' => array('error' => 'Not authenticated')));
+            return;
+        }
+                $disk_report = new Disk_report_model;
+                $out = array();
+                $out['stats'] = $disk_report->get_disk_type();
```
bochoven: What happens with $out?

tuxudo: That was an extra thing from testing. It has been removed.

PR #666, opened 2017-03-17
Hardware listing localization fix.

poundbangbash:


poundbangbash:
I think I did this wrong..found another way.  Will resubmit.

PR #670, opened 2017-03-19
Updated localadmin module

poundbangbash:
Added localization
Added Listing
Added Report
Updated model and controller.

Widget shows machines that has two or more local admins on the system. Since every computer has at least one admin account this shows those that have more than one.

Added widget name to the custom_defaults.php and alphabetized the current "small horizontal" listing.


poundbangbash:
Redacting.  Errors now showing on test system.

PR #672, opened 2017-03-20
Diskreport module fix

poundbangbash:
Links from the widget weren't working due to wrong module path name.  Changed path.
Updated widget title from "Volume Type" to "Storage Type" to better reflect data.


poundbangbash:
Cancelling and will resubmit.

PR #673, opened 2017-03-20
Fixed client tab localization.

poundbangbash:
Added missing entry in en.json for client tab.


poundbangbash:
Canceling and resubmitting to proper branch.

PR #675, opened 2017-03-20
Fix listings, tabs, and other small fixes

tuxudo:

PR #679, opened 2017-03-21
Localization fixes for directory service module

poundbangbash:
Localization fixes for directory service module


bochoven:
I think #675 also incorporated these changes

PR #684, opened 2017-03-22
Small bug fixes

tuxudo:
Rewrote Time Machine, Certificates, and Battery Condition widgets to use coloured blocks format instead of scrolling listing format


rickheil:
👍 awesome!

bochoven:
Please only submit things that are broken in the beta to the bug fix branch, otherwise the beta will go on forever! 😁

PR #686, opened 2017-03-23
Added clickable div to widgets. Links to appropriate listing.

poundbangbash:
Made widget title bars clickable to link to an appropriate listing.  This allows navigation to the listing without the buttons or scroll-box items filtering the view.


poundbangbash:
Further testing shows that the whole widget box is clickable.  Any buttons or listings clicked take precedence.

bochoven:
I'm not sure I understand what you're proposing. I think it is too implicit to just make the whole widget clickable. I would be ok with a little menu on the widget that would let you go to the corresponding listing.

poundbangbash:
Still playing around with this.
Moved the clickable area of the widget to just the titlebar instead of the whole widget.
Added a 15% invert for the hover highlight of the toolbar in the style.css file.

poundbangbash:
The PR is moot with the list-link changes made in wip.
Closing.

PR #706, opened 2017-03-29
Fixed media type column name localization.

poundbangbash:
Storage Report was missing a localization.
Added.  I don't have the translations for German or French so I just added it in English.


poundbangbash:
Removing to resubmit against bug-fix

PR #708, opened 2017-03-29
Removed extraneous parens.

poundbangbash:


poundbangbash:
Removed extraneous parens causing power listing errors.

PR #718, opened 2017-03-30
fix bug where ARD_AllLocalUsers key doesn't exist and causes crash

rickheil:
Thanks to @martinc on slack for reporting a bug where the ARD_AllLocalUsers key does not exist in some cases. This would cause a direct plist call to crash, whereas using the .get() method will set to none and move to the else block.


rickheil:
Moved to WIP... sorry!

PR #719, opened 2017-03-30
Added path to check for config, removing dependencey on a file in the…

poundbangbash:
… webroot. Allows for git management of custom config.php.

Since the common (preferred?) way to install munkireport is via git clone, this PR adds checking for a config.php in /custom/config/config.php as well. This will allow admins to manage a git repo on custom/config/ to version their preference file as well.

This first checks for config.php in /. If it finds it it loads and doesn't continue processing.  If the first check fails, it checks the /custom/config/ directory, if that fails then it errors out.

### DIFF
--- a/index.php
+++ b/index.php
@@ -25,11 +25,15 @@ function load_conf()
 	// Load default configuration
 	require_once(APP_ROOT . "config_default.php");
 
-	if ((include_once APP_ROOT . "config.php") !== 1)
-	{
+	if (file_exists(APP_ROOT . "config.php")) {
+		include_once(APP_ROOT . "config.php");
+	} elseif (file_exists(APP_ROOT . "/custom/config/" . "config.php")) {
+		include_once(APP_ROOT . "/custom/config/" . "config.php");
+	} else {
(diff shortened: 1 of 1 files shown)


bochoven:
I don't want to pin the location of an alternative config file to `APP_ROOT . "/custom/config/" . "config.php"`. And I am not sure what a sane alternative is.

poundbangbash:
How about a way to specify where the config should be similar to the App Settings in the config_defualt.php. Setting just the preference in the webroot config.php that directs where to find the full config.php.

/Config.php -> conf['config_path'] = "/admin/provided/path";

That's a bit messy to have a proxy config file but since the webroot location is the only default I don't see a way to set where the config should be found otherwise. 

I suppose a symlink would accomplish the same effect assuming the symlink would be followed. 

-Eric

bochoven:
You can accomplish the same by making `config.php` look like this: 

```php
<?php require '/admin/provided/path';
```

We could/should add that to the wiki as a way to securely store the config.

bochoven:
Added documentation here https://github.com/munkireport/munkireport-php/wiki/Secure-config.php

PR #721, opened 2017-03-31
Fix installpkg

rickheil:
Adds a line to install_script.php to touch the munkireport-$VERSION file - before, that was only done if you ran the curl | bash method. This helps because it means you can switch between deployment methods in the future.


rickheil:
Will re-base onto bugfix

PR #722, opened 2017-03-31
Touch version file when using pkg installer

rickheil:
Previous method only touched version file in SCRIPTDIR when using the
curl method to install.


bochoven:
Why not remove the touch part and only record the version in the plist?

rickheil:
Because I have an unnatural hatred for cfprefsd?

rickheil:
As discussed in slack, we should just use the plist version instead. Closing this one, will open a new one if new changes are needed. The touch file is a bit fragile.

PR #723, opened 2017-04-02
Certificate update

poundbangbash:
Re-worked the existing certificate module to gather info about keychains found in the System.keychain as well as /etc/certificates.  Keychains deployed via config profiles would appear in the System.keychain and keeping tabs on their expiration is valuable.  Originally this module said it was for servers, this makes it a module available to clients as well.

The data gathered now includes the issuer of the certificate and the path of where the cert is stored.

Updated listing and client tab to include the actual expiration date, issuer, and location.
Created a widget and added to report to show certificates grouped by name.


poundbangbash:
Wow. It was late. This is about getting data from certificates, not keychains as the original note references.

app/modules/certificate/locales/de.json
```diff
@@ -1,7 +1,12 @@
 {
+    "cert_groups": "Certificates by Name"
```
bochoven: It looks like a comma is missing?

app/modules/certificate/locales/fr.json
```diff
@@ -1,7 +1,12 @@
 {
+    "cert_groups": "Certificates by Name"
     "commonname": "Nom commun",
```
bochoven: comma?

app/modules/certificate/migrations/001_certificate_add_columns.php
```diff
@@ -0,0 +1,115 @@
...
+        $dbh = $this->getdbh();
+
+        switch ($this->get_driver()) {
+            case 'sqlite':
+                $dbh->beginTransaction();
+
+                // Create temporary table
+                $sql = "CREATE TABLE certificate (
+								id INTEGER PRIMARY KEY, 
+								serial_number VARCHAR(255) UNIQUE,
```
bochoven: I don't think serial_number should be unique

poundbangbash:
Thanks for pointing that out about the migration.  Not sure why it wasn't an issue in my testing.  I had multiple certs per SN.  Updated.

bochoven:
I think we need more indexes:

```php
        $this->rs['cert_path'] = ''; // Path to certificate
        $this->rs['cert_cn'] = ''; // Common name
```

also need an index, I think

PR #724, opened 2017-04-02
Fix for empty munkiinfo db

tuxudo:
Split munkiinfo listing from munkireport listing


bochoven:
It may be more informative to put all munkiinfo attributes in the munkiinfo listing table

tuxudo:
I don't know how to properly extract the data from the munkiinfo db because of how it's designed with only two columns.

bochoven:
I meant: just show the two columns..

PR #729, opened 2017-04-04
Fixed power model formatting

tuxudo:
Changed how nulls are detected


app/modules/power/power_model.php
```diff
@@ -359,12 +359,12 @@ public function process($data)
         }
 
         // Format voltage
-        if ( ! empty($this->cellvoltage)) {
+        if ( $this->voltage != "") {
              $this->voltage = ($this->voltage / 1000);
```
bochoven: Please convert voltage to int before doing arithmetic : http://php.net/manual/en/migration71.other-changes.php

PR #731, opened 2017-04-05
Update install_script.php

poundbangbash:
Stop prefpath creation when using the pkgbuild option.


poundbangbash:
Addressed in https://github.com/munkireport/munkireport-php/commit/e90cfde0fbf3816932dd9ec9e6f79eae536fcef3

PR #734, opened 2017-04-07
User Sessions module

tuxudo:
Jointly made by pudquick, clburlison, and tuxudo

### DIFF
--- a/app/modules/user_sessions/scripts/user_sessions.py
+++ b/app/modules/user_sessions/scripts/user_sessions.py
@@ -1,8 +1,8 @@
 #!/usr/bin/python
 """
 Parse user sessions on macOS so we can determine what users logged in and
-when the event took place. We only obtain 'console' sessions as 'ttys'
-sessions are less useful as a whole.
+when the event took place. We only obtain 'console' and 'ssh' sessions as
+regular 'ttys' sessions are less useful as a whole.
 
 Author: Clayton Burlison - https://clburlison.com
 
@@ -54,7 +54,16 @@ class utmpx(Structure):
                ]
 
 
-def fast_last(gui_only=True):
+def fast_last(session='gui_ssh'):
+    """This method will replicate the functionallity of the /usr/bin/last
+    command to output all logins, reboots, and shutdowns. We then calculate
+    the logout.
+
+    session takes on of the following strings:
+        * gui
+        * gui_ssh
+        * all
+    """
     # local constants
     setutxent_wtmp = c.setutxent_wtmp
     setutxent_wtmp.restype = None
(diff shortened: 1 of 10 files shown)


poundbangbash:
I loaded this branch up and took a look at the information. That's going to be a lot of info. With just two home computers I have 599 entries. Too much? 

I'd like to see a simplified view of unique usernames per machine -- just a list of all users that have logged into the computer.  That was my original intention with the FR.

How about making the timestamp in the listing more precise like the tab view?  If I'm looking for login info I'd be curious as to the exact times that that happened, not the general day that it happened.

tuxudo:
Mouseover the times in the listing for exact times.

poundbangbash:
I still think the specific time information is appropriate for this listing especially when there can be multiple entries per day.

app/modules/user_sessions/views/user_sessions_tab.php
```diff
@@ -0,0 +1,32 @@
...
+		<tr>
+            <th data-i18n="event"></th>
+            <th data-i18n="username"></th>
+            <th data-i18n="user_sessions.ipaddress"></th>
+            <th data-i18n="user_sessions.time"></th>
+		</tr>
+	</thead>
+	<tbody>
+<?php $user_sessionsitemobj = new user_sessions_model(); ?>
+      <?php foreach($user_sessionsitemobj->retrieveMany('serial_number=?', array($serial_number)) as $item): ?>
```
bochoven: Please use `retrieve_records` to make this business_unit aware

app/modules/user_sessions/views/user_sessions_tab.php
```diff
@@ -0,0 +1,32 @@
+<p>
+<table class="user_sessions table table-striped table-bordered">
```
bochoven: It would be nice to generate the table via a JSON call to the server instead of rendering php inline

app/modules/user_sessions/scripts/user_sessions.py
```diff
@@ -0,0 +1,134 @@
...
+            if e.ut_type == DEAD_PROCESS: event_label = 'logout'
+            if session is 'gui' and e.ut_line != "console":
+                continue
+            if (session is 'gui_ssh' and e.ut_host == "") and (
+                    session is 'gui_ssh' and e.ut_line != "console"):
+                continue
+            event = {'event': event_label,
+                     'user': e.ut_user,
+                     'time': e.ut_tv.tv_sec}
+            if e.ut_host != "":
```
bochoven: I sshd into my machine, but I did not get an entry..

app/modules/user_sessions/views/user_sessions_tab.php
bochoven: Please change this so I can merge this.

PR #737, opened 2017-04-10
Middleware reporting

rickheil:
Adds basic middleware reporting to munkiinfo module.

Still a WIP, and needs testing on machines that do not have middleware installed!


app/modules/munkiinfo/scripts/munkiinfo.py
```diff
@@ -96,6 +98,26 @@ def get_munkiprotocol():
     except AttributeError:
         return 'Could not obtain protocol'
 
+def middleware_checks():
+    """Check for middleware, get version if supported"""
+
+    middleware_version = []
+    name = []
+    for filename in os.listdir('/usr/local/munki'):
+            print "checking %s" % filename
```
clburlison: This debug statement needs to be removed

app/modules/munkiinfo/scripts/munkiinfo.py
```diff
@@ -96,6 +98,26 @@ def get_munkiprotocol():
...
+            if (filename.startswith('middleware') 
+                and os.path.splitext(filename)[1] == '.py'):
+                
+                name = os.path.splitext(filename)[0]
+                print 'middleware name %s' % name
```
clburlison: This debug statement needs to be removed

app/modules/munkiinfo/scripts/munkiinfo.py
```diff
@@ -106,9 +128,13 @@ def munkiinfo_report():
...
         AppleCatalogURL = ''
 
+    middleware_name, middleware_version = middleware_checks()
+
     report = {
         'AppleCatalogURL': AppleCatalogURL,
         'munkiprotocol': munkiprotocol,
+        'MiddlewareName': middleware_name,
+        'MiddlewareVersion': middleware_version,
     }
```
clburlison: This creates a string type when the s3 middleware is present
```xml
		<key>MiddlewareName</key>
		<string>middleware_s3</string>
		<key>MiddlewareVersion</key>
		<array>
		</array>
```

However it creates a double array when no middleware is present
```xml
		<key>MiddlewareName</key>
		<array>
		</array>
		<key>MiddlewareVersion</key>
		<array>
		</array>
```

This needs to be corrected to be an array of strings all of the time.

app/modules/munkiinfo/scripts/munkiinfo.py
```diff
@@ -96,6 +98,26 @@ def get_munkiprotocol():
...
+            if (filename.startswith('middleware') 
+                and os.path.splitext(filename)[1] == '.py'):
+                
+                name = os.path.splitext(filename)[0]
+                print 'middleware name %s' % name
+                try:
+                    middleware = importlib.import_module(name)
+                    middleware_version = middleware.get_version()
+                except:
+                    print "Error importing middleware for version checks, or version function does not exist."
```
clburlison: This debug statement is of value however also should likely be removed and replaced with a `pass`

To my knowledge @waderobson is the only person in the community that has publicly release middleware for munki. So I'm assuming this `get_version()` method is for your own custom middleware. It might be of value to ping the [s3 auth](https://github.com/waderobson/s3-auth/) & [gcs-auth](https://github.com/waderobson/gcs-auth) middleware and ask Wade to add a `get_version()` method (or possibly create a pull request).

clburlison: On second thought having this statement print when the verbose flag is passed via `managedsoftware` might be nice. I would add more info at the beginning of the statement 
```bash
print "ERROR: munkiinfo.py - Error importing middleware for version checks, or version function does not exist
```

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: Good find on the double array, added to my todo

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: From what I've seen so far (and it hasn't been much because middleware isn't heavily used yet), Wade's code is the reference implementation but most admins who are using it seem to be rolling their own at least to some degree - hence the update to the readme flagging the need to provide the get_version() function. I'll chat with Wade separately and see if he has interest in PRs for this. It would be a good idea to also add it to the munki documentation if Wade is open to PRs

rickheil: Discussed with Wade on slack - he is open to PRs. I'll file those and update docs this week.

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: As we discussed on slack, it may be cleaner to do something like MiddlewareName_Version and do it in a single string other than adding two strings.

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: Removed

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: I split the except into two - one for ImportError (something wrong with naming or syntax in the file), and one for AttributeError (which will catch middleware that imports properly but doesn't have a __version__ global variable).

app/modules/munkiinfo/scripts/munkiinfo.py
rickheil: As discussed on slack we're doing Name-Version, so something like middleware_s3-1.0 will show up. I beefed up the logic to return a blank string when the preference isn't read properly.

PR #750, opened 2017-04-13
Change power.schedule to TEXT

tuxudo:


app/modules/power/migrations/004_power_schedule_text.php
```diff
@@ -0,0 +1,55 @@
...
+                    $this->exec($sql);
+                
+                    // Delete old table
+                    $sql = 'DROP TABLE power_tmp';
+                    $this->exec($sql);
+
```
bochoven: You need to add indexes

tuxudo: I'm not sure what the proper method is for adding indexes via a migration.

bochoven: Check https://github.com/munkireport/munkireport-php/blob/master/app/modules/machine/migrations/002_machine_add_indexes.php

app/modules/power/migrations/004_power_schedule_text.php
```diff
@@ -0,0 +1,73 @@
...
+                    $sql = 'DROP TABLE power_tmp';
+                    $this->exec($sql);
+                
+                // Get database handle
+                $dbh = $this->getdbh();
+
+                $dbh->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
+
+                // Wrap in transaction
+                $dbh->beginTransaction();
```
bochoven: Please move beginTransaction to the start of the sqlite section. Transactions make the migration atomic.

PR #752, opened 2017-04-14
Text-muted causes some text to be unable to be seen.

WardsParadox:
In some dark themes, applying the text-mute attribute results in being unable to see the text at all. I haven't
screencast 2017-04-14 08-48-20


bochoven:
I only see this problem in the Superhero theme. It seems that we should fix the problem there. Maybe we can reach out to the developer for a solution.

PR #753, opened 2017-04-17
Example missing ;

poundbangbash:
Example Inventory path ignore was missing a ; leading to failures when copy and pasting into the config.php.


bochoven:
Would you mind sending the PR against the bug fix branch or wip?

bochoven:
Fixed in #754

PR #771, opened 2017-05-07
Add LetsEncrypt specific path check

poundbangbash:
LetsEncrypt saves its certs in /etc/letsencrypt/live/<domain> by default.
This adds the check for certs there.


bochoven:
Would it be possible to add certpaths to an array and loop over them? Maybe there are still more interesting cert locations to be added.

poundbangbash:
Updated script to cycle thru an array of paths for scalability.

The System Keychain method still needs to be separate since it uses different commands to parse out the cert info.

PR #786, opened 2017-05-15
Add event to list of default modules

gmarnin:


bochoven:
Already fixed here https://github.com/munkireport/munkireport-php/commit/f79d6faf7f3339cbb1ce1ed9240e40089519d4a9

PR #798, opened 2017-05-17
Create fr.json

lolipale:
Warning : some duplicate keys
tm0p_short, tm0p, ti0p, ti1p, tm0p2, tc0f, tc0h, tc0j


bochoven:
@tuxudo There are some duplicate keys in the en.json, could you remove those? Apparently they have different strings attached.

PR #811, opened 2017-05-19
Remove index.php? from appUrl for cleaner URLs

cwhits:
This one's a super tiny change but makes use of the `mod_rewrite` rule already in the supplied `.htaccess` file.

Prior to this change, pressing the `d` hotkey would send you to your `baseUrl` + `index.php?/show/dashboard`. This is makes it so all the urls are cleaner and don't include `index.php?`.’

Any external services hard-coded to use `index.php?`-style urls will still function just fine since mod_rewrite won't touch them.

### DIFF
--- a/app/views/partials/head.php
+++ b/app/views/partials/head.php
@@ -34,7 +34,7 @@
 
 	<script>
 		var baseUrl = "<?php echo conf('subdirectory'); ?>",
-			appUrl = baseUrl + 'index.php?',
+			appUrl = baseUrl.replace(/\/$/,''),
 			businessUnitsEnabled = <?php echo conf('enable_business_units') ? 'true' : 'false'; ?>;
 			isAdmin = <?php echo $_SESSION['role'] == 'admin' ? 'true' : 'false'; ?>;
 			isManager = <?php echo $_SESSION['role'] == 'manager' ? 'true' : 'false'; ?>;


cwhits:
It'll work as long as they enable a rewrite rule, which both support. Maybe we find someone who uses either and add it to the readme?

bochoven:
Not everyone is able to use rewrites, I would not change the default. You can, however override the default in config.php. Hotkeys should honour this setting

cwhits:
Hotkeys _aren't_ honoring this right now because they use `appUrl`, and `appUrl` is set in `head.php` as 
```javascript
var baseUrl = "<?php echo conf('subdirectory'); ?>",
    appUrl = baseUrl + 'index.php?',
```

I'll see if I can adjust `appUrl` to use `conf('index_page')` or if it's blank to use baseUrl with the trailing slash trimmed.

bochoven:
AppUrl was meant for API calls, I should not have used it for the hotkeys, they should obey the config settings.

cwhits:
`appUrl` is used in a lot more than just the hotkeys—it's used petty much everywhere something is generated with JavaScript. Each link on the Client Report takes you to `index.php?/clients/detail/serialnumber`, URLs for all the widgets on the dashboard, Admin > Show System Status...

Are you opposed to changing `appUrl` to `appUrl = <?php echo conf('index_page') ? 'baseUrl + '.conf('index_page') : 'baseUrl.replace(/\/$/,\'\')' ?>;`?

bochoven:
Closing as this should be fixed here: a754b2a06c2e737ccf223ffca37688d62bfe4dcf

PR #814, opened 2017-05-22
Firmware listing - differentiate between "No" and no data

### LINKED ISSUE #813: Firmware password status not correct
I’ve noticed that in listing - security, the Firmware Password status is set to Disabled by default. So clients who have not updated their MR yet, are reflected as ‘Disabled’. The Firmware _widget_ is working fine though.

So, the status should be set to unknown or stay blank.

rickheil:
Fixes #813


bochoven:
Maybe you can fix the current behaviour in wip?

rickheil:
Must have made this on an out of date copy of WIP then, I'll resolve the conflicts later today

rickheil:
Your commit to wip already fixed this issue so no further action needed. I still need to look into the 'command' value being in DB for firmwarepw but will open another PR for that.

PR #831, opened 2017-06-23
Move firewall column

tuxudo:


bochoven:
Any motivation for this change?

tuxudo:
It's purely a UI change to keep all the red/green columns together.

PR #834, opened 2017-06-23
Highest Supported OS module

tuxudo:
Lists the highest supported OS version for a machine


app/modules/munkireportinfo/munkireportinfo_model.php
```diff
@@ -56,7 +56,6 @@ public function process($data)
             // Convert version to int
             if (isset($plist['version'])) {
                 $digits = explode('.', $plist['version']);
-                array_pop($digits);
```
bochoven: why is this fix in the PR?

bochoven:
Where are the conditions coming from? They look far more simple than the comparison thing @ hjuutilainen set up here:

https://github.com/hjuutilainen/adminscripts/blob/master/check-10.12-sierra-compatibility.py

I think the whole conditions part should be moved to config.php in an array of some sorts. This should help people that cannot update munkireport, but need to report on newer models.

app/modules/munkireportinfo/munkireportinfo_model.php
tuxudo: It was a bug discovered with the MRinfo module. Wasn't supposed to be in this PR. Will remove change and make PR for it.

tuxudo:
The conditions are all within the model. It takes the machine ID, iMac7,1, and turns it into a string "iMac71". It then checks for the machine family, iMac, and removes the machine family from the string. It then checks the resulting int, 71, to see if it is in the range of supported OSes starting with the highest and working down. 

I thought this method would be better than hard coding in every single Mac model ID or logicboard ID. If a new machine, such as iMac20,1, was to check it it would fall into the range that would set its highest OS as the most recent.

bochoven:
I like the idea, but not so sure on the implementation 😄 
At the moment, maintaining this would require quite some effort. It would be easier if the module could pull the data from some external source (or even an internal source if that exists).
If that's not possible, I would change the whole 'if then' comparison thing into a big array like:

```
$max_version = [
    'iMac' => [
        '71' => '10.11.6',
    ]
];
```

And process that, so that it's easy for people to add new datapoints.

tuxudo:
I don't know how to do the array thing without hard coding in every single known Mac model. I'd prefer to have the data source within MR to prevent the module from breaking if the external source changes or goes offline.

PR #844, opened 2017-07-14
When enabling hide_inactive_modules some sections appear broken on the Client Summary page.

chrisgrande:
When using hide_inactive_modules = TRUE a few sections (Time Machine, ARD, Security, etc) on the client summary page appear broken, with missing headers, content, etc. unless their modules are enabled. This minor fix will now hide these sections unless they are specified in the modules section with hide_inactive_modules enabled or show them all if hide_inactive_modules is disabled.


bochoven:
I agree that this should be handled better, but I think it needs to be addressed on a higher level and not in the template. Please leave this PR open as a reminder though.

bochoven:
no better ideas came to mind.. c.losing

PR #849, opened 2017-08-02
Update plist buddy for case-sensitivity

GordSpence:
Enabling compatibility with case-sensitive file systems


bochoven:
Please use the wip branch https://github.com/munkireport/munkireport-php#contributing

PR #850, opened 2017-08-03
Module for admin-provided munki conditions

### LINKED ISSUE #627: Module to track admin-provided conditional items
In Slack today, we were discussing a possible module to track if a particular Mac can run 10.12 Sierra - and I realized we already deploy [this script](https://github.com/hjuutilainen/adminscripts/blob/master/check-10.12-sierra-compatibility.py) from Hannes Juutilainen.  Thus, that data is already recorded in /Library/Managed Installs/ManagedInstallReport.plist (though it doesn't display in Additional Munki Info, so I don't think it's in the MunkiReport database yet).

If there were a way to view, and sort, [admin-provided conditional items](https://github.com/munki/munki/wiki/Conditional-Items#admin-provided-conditions), that would take care of this particular request and possibly sidestep the need for many custom modules.  It would add expansion possibilities in the same way that the ARD module can be used to display custom data.

Do you think this would be possible to add?

Obviously up to you guys, but I'm thinking it could look like the Installation Ratio report, or the Inventory listing, where you can select individual items and view more information about them.  That way, the module would be able to support multiple conditions with multiple values (such as yes or no, or an array with more than two values).

Thanks for considering this!

### LINKED ISSUE #835: FR: display custom Munki conditions
It'd be really awesome if MunkiReport could display all of the admin-provided conditions in `/Library/Managed Installs/ManagedInstallReport.plist`, as detailed [here](https://github.com/munki/munki/wiki/Conditional-Items).

An example use case: we currently deploy @hjuutilainen's [script](https://github.com/hjuutilainen/adminscripts) for checking if a Mac can run macOS 10.12.  This leaves a `sierra_supported` custom condition on disk that can be used to determine if we can deploy a 10.12 installer to that Mac.  This data is not currently retrievable in MunkiReport.

My thinking is that if we wanted to retrieve data about each Mac, we can just deploy scripts and have MunkiReport retrieve those conditions - this might reduce the need for new modules.

nathanperkins:
resolve #627 from @flammable.
resolve #835 from @flammable.

Please let me know if you have any suggestions.

For one, I didn't know how to get the location of ManagedInstallReport.plist from prefs so it is hardcoded in the client script.

Second, all values are truncated and stored as 255 character strings. Would it be preferred to utilize BLOB so that JSON could be used to store arrays?

Munki Conditions Module
=======================

Provides the results of [munki conditions](https://github.com/munki/munki/wiki/Conditional-Items), allowing admins to easily generate their own reportable datapoints without creating custom MunkiReport modules. Condition scripts are even easier to create using [munki-facts](https://github.com/munki/munki-facts).

Admin-Provided Condition scripts are found in `/usr/local/munki/conditions`.
The results of these scripts are stored in the 'Conditions' dictionary in `/Library/Managed Installs/ManagedInstallReport.plist`.

Both keys and values are truncated to 255 characters.

The following information is stored in the munki_conditions table:

* serial_number
* condition_key
* condition_value

This module is mostly based on work by clburlison, erikng, bochoven and more.

screen shot 2017-08-03 at 12 12 40 am
screen shot 2017-08-03 at 12 12 55 am

### DIFF
--- a/app/modules/munki_conditions/README.md
+++ b/app/modules/munki_conditions/README.md
@@ -1,0 +1,14 @@
+Munki Conditions Module
+=======================
+
+Provides the results of [munki conditions](https://github.com/munki/munki/wiki/Conditional-Items), allowing admins to easily generate their own reportable datapoints without creating MunkiReport modules. Condition scripts are even easier to create using [munki-facts](https://github.com/munki/munki-facts).
+
+Admin-Provided Condition scripts are found in `/usr/local/munki/conditions`.
+The results of these scripts are stored in the 'conditions' dictionary in `/Library/Managed Installs/ManagedInstallReport.plist`
+
+The following information is stored in the munki_conditions table:
+
+For each entry in conditions:
+
+* condition_key
+* condition_value
--- a/app/modules/munki_conditions/README.md
+++ b/app/modules/munki_conditions/README.md
@@ -1,14 +1,17 @@
 Munki Conditions Module
 =======================
 
-Provides the results of [munki conditions](https://github.com/munki/munki/wiki/Conditional-Items), allowing admins to easily generate their own reportable datapoints without creating MunkiReport modules. Condition scripts are even easier to create using [munki-facts](https://github.com/munki/munki-facts).
+Provides the results of [munki conditions](https://github.com/munki/munki/wiki/Conditional-Items), allowing admins to easily generate their own reportable datapoints without creating custom MunkiReport modules. Condition scripts are even easier to create using [munki-facts](https://github.com/munki/munki-facts).
 
 Admin-Provided Condition scripts are found in `/usr/local/munki/conditions`.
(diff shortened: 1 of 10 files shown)


rickheil:
The munkireport module grabs the Munki installation dir this way:

`managed_install_dir = CoreFoundation.CFPreferencesCopyAppValue(
    "ManagedInstallDir", "ManagedInstalls")`

nathanperkins:
Thanks, I made some adjustments to the script to use the munki prefs, with a fallback to a default path if the pref doesn't work.

nathanperkins:
Hmm, I need to chase down a bug where the condition data is overwriting other data in the client tabs.

nathanperkins:
@tuxudo suggests that the values be stored as BLOB instead of VARCHAR255. I will work on that next.

bochoven:
Sorry for the late reaction. This looks great. Just one nitpick: could you rename the module to munki_facts? Although they are used as conditions, I think they should be treated as facts.

PR #851, opened 2017-08-03
PlistBuddy Fix

GordSpence:
Updating PlistBuddy usage so that it is compatible with case-sensitive FileSystems.


bochoven:
Excellent, thanks!

PR #855, opened 2017-08-09
Added more deafult filters for inventory module

poundbangbash:
Added filters that are useful to reduce the noise in the inventory module.


bochoven:
Thanks, could you fix the spacing as well? And add a trailing comma to the last item in `bundleid_ignorelist`?

poundbangbash:
Updated.

PR #867, opened 2017-09-11
Different suffix for pre-/postflight scripts

gniemetz:
I have a master-pre-/postflight script, which calls the munkireport and other pre-/postflight scripts. When i install munkireport on the client my master-scripts are being overwritten, so i needed an option to create those pre-/postflight scripts with different suffix. This is optional and can be set through the config.php

### DIFF
--- a/app/views/install/install_script.php
+++ b/app/views/install/install_script.php
@@ -11,6 +11,8 @@ PREFLIGHT=1
 PREF_CMDS=( ) # Pref commands array
 TARGET_VOLUME=''
 CURL=("<?php echo implode('" "', conf('curl_cmd'))?>")
+PREFLIGHT_SUFFIX="<?php echo conf('preflight_suffix'); ?>"
+POSTFLIGHT_SUFFIX="<?php echo conf('postflight_suffix'); ?>"
 # Exit status
 ERR=0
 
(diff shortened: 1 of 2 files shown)


bochoven:
First, you could put your master scripts in preflight.d and postflight.d.

Second, if you want to pursue this, I think it would be less confusing to just put the name of the target scripts in `config_default.php`:

```php
$conf['preflight_script'] = 'preflight';
$conf['postflight_script'] = 'postflight';
$conf['report_broken_client_script'] = 'report_broken_client';
```

So people can override the name with wathever they like.

gniemetz:
Ok, my idea was to leave the names of pre-/postflight the same, but add an optional suffix if needed, configurable via config.php with parameter $conf['preflight_suffix'] and $conf['postflight_suffix']

bochoven:
Please just define the full names, suffix is confusing for other people. It suggests more than just changing the name. The preferred method is to put your custom scripts in pre|postflight.d, why not use that?

gniemetz:
I make use of https://github.com/munki/munki/wiki/WPKG-like-Dynamic-Manifests-Without-CGI and https://github.com/munki/munki/wiki/Dynamic-Client-Configurations-Via-Preflight-Scripting which are started from my own preflight script. This script also starts the preflight of munkireport.
Would it be ok to close this pull request and propose a new one as stated in https://github.com/munkireport/munkireport-php/pull/867#issuecomment-328451004?
Thank you

bochoven:
Sure!

PR #868, opened 2017-09-15
Assign TPL_BASE after options read

fuzzylogiq:
Otherwise it is always set to contain BASE_URL defined at top of script and cannot be overridden.


bochoven:
I'm afraid I don't get this. Options are read before the template is applied.

fuzzylogiq:
If, for instance, I run the script with the -b option to override the BASEURL, TPL_BASE is still set to whatever the php code has pulled from the conf as the BASEURL (as it is set at the top and not re-set after the -b option is read with getopts)

bochoven:
I tried to change the base branch, maybe you can resubmit against wip?

PR #876, opened 2017-09-26
Add SKEL tracking to security module

rickheil:
Add secure kernel extension loading ("SKEL") tracking to the security module, supporting 10.13 and up. Reports enabled/disabled for 10.3, reports "not supported" for others. Added to listing and to a widget.


rickheil:
Also added to client summary tab and added the necessary JS.

rickheil:
Also - if you think this would fit better as a direct part of the extensions module, I'm happy to move it over there. LMK.

rickheil:
After some discussions on Slack, I've changed the naming and behavior a bit. Machines older than 10.13 or where User Approved Kernel Extension Loading (aka UAKEL aka SKEL) is in the disabled state will show as "Open" (security.skel.all-approved). Machines with UAKEL/SKEL turned on in the default state will show up as "User Approved" (security.skel.user-approved). I've made both widget buttons blue because I don't think either configuration is necessarily deserving of either green or red. Open to further feedback and corrections. I think all the bugs are gone so please test!

PR #884, opened 2017-11-03
Hardware age SQL fix

BooitsDaly:
SQL statement for determining hardware age now excludes machines that come back as NULL


app/modules/warranty/warranty_controller.php
```diff
@@ -108,13 +108,13 @@ public function age()
         $where = get_machine_group_filter();
 
         $sql = "SELECT count(1) as count, 
-				$agesql AS age 
-				FROM warranty
-				LEFT JOIN reportdata USING (serial_number)
-				$where
```
cwhits: I think this needs to be added back in.

cwhits: _Old_ line 114 specifically.

bochoven:
Thanks for the PR! I think the proper SQL syntax should be 

```
AND $agesql IS NOT NULL
```

PR #886, opened 2017-11-06
Added filesystem widget

tuxudo:
Requested by @gmarnin

### DIFF
--- a/app/modules/disk_report/disk_report_model.php
+++ b/app/modules/disk_report/disk_report_model.php
@@ -80,11 +80,12 @@ class Disk_report_model extends \Model
      **/
     public function get_volume_type()
     {
-        $sql = "SELECT VolumeType as type, count(*) as count
+        $sql = "SELECT COUNT(CASE WHEN VolumeType = 'APFS' THEN 1 END) AS apfs,
+						COUNT(CASE WHEN VolumeType = 'bootcamp' THEN 1 END) AS bootcamp,
+						COUNT(CASE WHEN VolumeType = 'Journaled HFS+' THEN 1 END) AS hfs
 						FROM diskreport
 						LEFT JOIN reportdata USING (serial_number)
 						WHERE Internal = 1
-						GROUP by VolumeType
 						".get_machine_group_filter('AND');
         return current($this->query($sql));
     }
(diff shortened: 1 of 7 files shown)


app/modules/disk_report/disk_report_model.php
```diff
@@ -72,6 +72,23 @@ public function get_disk_type()
...
     }
+    
+    /**
+     * Get filesystem type statistics
+     * autor tuxudo
+     *
+     **/
+    public function get_volume_type()
+    {
+        $sql = "SELECT COUNT(CASE WHEN VolumeType = 'bootcamp' THEN 1 END) AS bootcamp,
```
bochoven: It may be better/more future proof to do something like ```SELECT VolumeType as type, count(*) as count from diskreport LEFT JOIN reportdata USING (serial_number) WHERE Internal = 1 group by VolumeType```

bochoven:
Maybe you could also add a filesystem report that features the widget?

tuxudo:
Added filesystem widget to storage report

app/modules/disk_report/disk_report_model.php
```diff
@@ -72,7 +72,23 @@ public function get_disk_type()
...
-
+    
+    /**
+     * Get filesystem type statistics
+     * autor tuxudo
+     *
+     **/
+    public function get_volume_type()
+    {
+        $sql = "SELECT VolumeType as type, count(*) as count FROM diskreport
```
bochoven: There's two times `FROM diskreport` in the query, does that even work? And please fix the indenting.

tuxudo:
Requested changes made.

bochoven:
Please test this PR as the widget is not working here.

tuxudo:
Issues resolved, tested latest commits.

app/modules/disk_report/views/filesystem_type_widget.php
```diff
@@ -0,0 +1,39 @@
+<div class="col-lg-4 col-md-6">
+	<div class="panel panel-default" id="filesystem-type-widget">
+		<div class="panel-heading" data-container="body">
+			<h3 class="panel-title"><i class="fa fa-hdd-o"></i>
+			    <span data-i18n="disk_report.filesystem_type"></span>
+			    <list-link data-url="/show/listing/disk_report/disk_report"></list-link>
```
bochoven: This link is not correct

Issue #892, opened 2017-11-14
smart module hangs when creating munkireport installer


sphen13:
been having this for awhile.  maybe its just me - but when building the munkireport package for munki it hangs at the smart module.  i check and its actually hung on this process:
`update-smart-drivedb`

when i manually run it i get:
```
$ /bin/sh /usr/local/sbin/update-smart-drivedb
override rw-r--r--  root/admin for /usr/local/share/smartmontools/drivedb.h.old? (y/n [n]) y
/usr/local/share/smartmontools/drivedb.h updated from branches/RELEASE_6_5_DRIVEDB
```

it is asking for confirmation - do you think there is any risk in passing a `y` to this automatically?

PR #894, opened 2017-11-15
Fix smart_stats update-smart-drivedb interactive prompt

sphen13:
PR to address issue #892

### DIFF
--- a/app/modules/smart_stats/scripts/install.sh
+++ b/app/modules/smart_stats/scripts/install.sh
@@ -18,9 +18,9 @@ else
     
 	# Update smartctl database
 	if [[ -f /usr/local/sbin/update-smart-drivedb ]]; then
-	     /usr/local/sbin/update-smart-drivedb > /dev/null 2>&1
+	     /usr/bin/yes | /usr/local/sbin/update-smart-drivedb > /dev/null 2>&1
 	elif [[ -f /usr/local/bin/update-smart-drivedb ]]; then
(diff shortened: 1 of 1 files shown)


sphen13:
well - honestly i am not well versed in smartmontools etc - not sure why the drivedb needs to be updated etc.  was just trying to fix a hang when the install.sh is run to build the pkg and modules that i have been seeing.  

looks like it also depends on gpg to ensure update consistency.. which you dont have :) in your case i would think install.sh would just error that line and continue - not hang.  we could add --no-verify too though.

maybe this is a better question for @tuxudo who initially set up this module?  thx! :)

tuxudo:
The reasoning for having it update the smartdb is to make sure that smartctl always has the latest drives in its database. Not everyone runs the latest version of smartmontools and the online database is updated more often than a new version of smartmontools is released. It's possible to skip the updates, but at the possible cost of the smartctl not properly displaying stats for more modern or recently released drives. Some notable drives that are updated or included in the database often are the PCIe blade SSDs included in Macs.

sphen13:
hey @tuxudo - makes sense... so either way not exactly sure what we should do here.  i think we should make the change that i had in here to make sure the database is updated and permissions differ - but interestingly @bochoven indicated a different issue which looks like it was just fixed in this commit (path issue):
https://github.com/mirror/smartmontools/commit/cb47b33b56c124b70e1d884d6f7e1caead1ff880

thoughts?

PR #896, opened 2017-11-21
Update Network Info script to now pull Tunnel adapter information

jbaker10:
This is a new python script which will replace the current shell script. It behaves very much the same way, no backend changes are needed to implement this.

The biggest addition is that this script will pull tunnel information for any "utun" adapter (except utun0 as I believe that is used for internal services like Back to My Mac).


bochoven:
Looks fine. Please add removal of the old .sh script in both install.sh and uninstall.sh (in case of upgrading)

bochoven:
Oh, and you should sent your PR to wip and not to master

PR #897, opened 2017-11-21
Update Network Info script to now pull Tunnel adapter information

jbaker10:
This is a new python script which will replace the current shell script. It behaves very much the same way, no backend changes are needed to implement this.

The biggest addition is that this script will pull tunnel information for any "utun" adapter (except utun0 as I believe that is used for internal services like Back to My Mac).


app/modules/network/scripts/install.sh
```diff
@@ -4,17 +4,17 @@
 NW_CTL="${BASEURL}index.php?/module/network/"
```
bochoven: Please move `rm -f ${MUNKIPATH}preflight.d/networkinfo.sh` here.

PR #882, opened 2017-10-25
Sort "Set Filters" by name

BooitsDaly:
Fixes #740

PR #898, opened 2017-11-21
Sort machine groups by name

choules:
Having more than a few machine groups leads to an confusing filter view because the list was sorted by id and not by name. This pull request now sorts this list by group name.


rickheil:
#882 covers this as well

choules:
Absolutely right, must have missed that.

PR #901, opened 2017-11-27
Moved around power/battery tab components

tuxudo:
Moves UPS devices into battery tab
Adds battery/UPS percent as client tab badge
Fixes alignment with UPS
Moves and hides adapter block in power client tab


tuxudo:
Closing PR, changes rolled into PR #926 to fix some problems with UI logic

PR #905, opened 2017-11-29
Updated Homebrew modules

tuxudo:
Added more keys and changed data types


bochoven:
If serial_number is not unique, please add an index for it.

PR #906, opened 2017-11-29
Remove unused Servermetrics.sh script

tuxudo:
Removed obsolete Servermetrics.sh script from module.


tuxudo:
I think this script is still used. Not actually sure as the Server Metrics client tab no longer appears, but I think it should. Maybe this should be closed and not merged.

PR #907, opened 2017-11-30
Added support for 10.13+ caching server

tuxudo:
Basically a whole new module bolted on to the existing caching module. Can be installed on all 10.13 clients to get data about reachable caching servers.

Includes: 
- Many more widgets
- New listing and config options
- Data set aware client tab
- Expanded caching report


tuxudo:
I have some more ideas to rework this further. It's working how it is, but is missing some functionality that the older 10.8-10.12 caching server has that I'm going to try to work into it.

tuxudo:
I can't get the module to do what I want without really messing with the kiss core thingy. Maybe another day. Module can be merged now.

PR #909, opened 2017-12-01
Add Database Size

gmarnin:
This PR adds the mysql database size in MB to the database column on the system status page.

### DIFF
--- a/app/controllers/system.php
+++ b/app/controllers/system.php
@@ -32,6 +32,7 @@ class system extends Controller
             'db.driver' => '',
             'db.connectable' => false,
             'db.writable' => false,
+            'db.total_size' => '',
             'error' => '',
         );
         $config = array(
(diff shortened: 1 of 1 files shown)


app/lib/munkireport/Database.php
```diff
@@ -98,4 +98,27 @@ public function get_driver()
...
+    
+	/**
+     * Get database size
+     *
+     * @return number
+     **/
+    public function get_totalsize()
+    {
+    
+    	$sql = 'SELECT sum(round(((DATA_LENGTH + INDEX_LENGTH) / 1024 / 1024 ), 2)) from information_schema.TABLES';
```
bochoven: This only works for MySQL and I think it also needs a WHERE TABLE_SCHEMA = 'insert munkireport table from config'

app/helpers/site_helper.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 // Munkireport version (last number is number of commits)
-$GLOBALS['version'] = '2.15.0.2795';
```
bochoven: This should not be part of the commit

bochoven:
Please check if #1094 works for you

PR #926, opened 2017-12-20
Made power.sh into Python

tuxudo:
@bochoven rejoices in hoping that @tuxudo has finally embraced Python.


bochoven:
Please use `system_profiler -xml SPPowerDataType` and parse the result with `plistlib` (which delivers structured data for easy parsing)

rickheil:
Also worth noting ioreg supports xml output.

tuxudo:
This should all fix issue #929

app/modules/power/scripts/power
```diff
@@ -0,0 +1,376 @@
...
+                batteryprofiler['SerialNumber'] = obj[item]
+            elif item == 'sppower_ups_installed':
+                batteryprofiler['UPSInstalled'] = obj[item]
+            elif item == 'sppower_battery_installed' and obj['sppower_battery_installed'] == "TRUE":
+                batteryprofiler['condition'] = obj['sppower_battery_health_info']['sppower_battery_health']
+            elif item == 'sppower_battery_installed' and obj['sppower_battery_installed'] == "FALSE":
+                batteryprofiler['condition'] = "No Battery"
+            elif item == 'sppower_battery_charge_info':
+                try:
+                    batteryprofiler['MaxCapacity'] = obj[item]['sppower_battery_max_capacity']
```
bochoven: If you are trying to get a property of a dictionary and you're not sure that it exists, use get(). So in this case: 

```python
batteryprofiler['MaxCapacity'] = obj[item].get('sppower_battery_max_capacity', null)
```

https://www.programiz.com/python-programming/methods/dictionary/get

tuxudo: That does not have the desired result of not having the key at all in the resulting plist.

bochoven: Why do you set the error to an empty string?

tuxudo: That line is there just to give Python something to do. The error isn't used in the model

app/modules/power/scripts/power
```diff
@@ -0,0 +1,380 @@
...
+import plistlib
+import os
+import sys
+import string
+import re
+
+def get_battery_profiler():
+    '''Uses system profiler to get power info for this machine.'''
+    cmd = ['/usr/sbin/system_profiler', 'SPPowerDataType', '-xml']
+    proc = subprocess.Popen(cmd, shell=False, bufsize=-1,
```
mosen: depending on the information you want to get from subprocess, it can be cleaner to use the `check_output` method which only cares about the output of the command you are running.

tuxudo:
Closing PR, module moved to its own repo

PR #933, opened 2017-12-23
Revamped Fans and Temperatures Module

tuxudo:
Reports back data about amps, voltages, watts, and SMC. Written in
Python 💃 and better error handling.


ofirgalcon:
Had problems building the client installer after installing these. I gave up as I'm still setting up my server.

Issue #102, opened 2014-05-12
email helper


ghost:
I'd like to include an email helper so that we can configure email alerts for some modules.

For example I have in mind these alerts:
- Alert for each new computer. Include serial, mac, IP ...
- When a stolen/lost computer/display is seen
- When a computer reports SMART failures
- Broken client

diwanicki:
any progress on this? Would be nice to get alerts for errors and warnings

childrss:
+100 -- we're a big University and I would like to set a notification e-mail on a per Business Unit level for all machines registering into that business unit. 

Adding an e-mail field for the business unit would allow for the potential of adding alerting/reporting for that unit to go into just a regular e-mail box, OR better yet, a helpdesk ticketing system for follow up!

We have a campus unit that handles all of our Apple sales and repair work, and many times our Faculty skip their local IT techs and go right to the retail sales/repair store.  Adding a Business Unit section that held the local unit's helpdesk e-mail field (and a phone number contact field) as well as perhaps a "person contact phone number / e-mail field"  would be extremely helpful to campus-level repair shops.

Other Useful Examples:  
1. alerting on either TimeMachine/Crashplan backups at X number of days without a backup triggers an e-mail to the appropriate business unit's helpdesk ticketing system, and they followup with the customer.
2. and in the future an alert from your AntiVirus/Malware scanner going directly to your helpdesk ticketing system for followup with the person you support...

As a front-line IT guy I'm trying to wrangle all the University units into an idea of cohesive mac support for the people we support  "from cradle to grave" (from when they open the box with Apple's DEP-> MDM-> EPM, to getting it repaired as needed, and then decomissioning)  to give the people we support superior experience and that means we've got to communicate better, and MR is gonna play a BIG part in that: the other reporting systems I've seen are no where near as easy to setup, use and have such immediate return on investment in setting up (we discovered a hard drive in a SMART failing status the first time we pushed out munkireports, and it converted our support immediately from REactive to PROactive)

tuxudo:
This should be addressed with PR #935

macvfx:
yes, please 💯 email and / or notifications

PR #935, opened 2017-12-29
Add email notifications to MR events and actions

tuxudo:
Uses PHPMailer and is currently triggered by the same things that trigger an event with support for more as requested or needed. Should resolve Issue #102


bochoven:
Although I appreciate the effort, I cannot merge this. Emailing needs to have a couple of things that are missing:

- emails should be queued via a queueing system like redis, beanstalk, etc. preferably in an independent way (make it possible to switch)
- we need to be able to create digest emails that contain mulitlple messages that are sent on specific times
- we need to be able to create filters on emails
- it would even be better to have a notification system that can send emails, text messages, push notifications, etc.

tuxudo:
Closing PR, it was a fun exercise

PR #936, opened 2018-01-01
Converted smart_stats script into Python

tuxudo:
It finally happened. SMART Stats' script is now Python and way faster. (@bochoven is thrilled)


bochoven:
@tuxudo I fixed the old smart stats installer (you cannot initialise anything on install.sh as it can also run on an autopkg machine when it builds a package). Maybe you can incorporate the changes in this PR?

tuxudo:
Closing PR, moved to module repo

PR #938, opened 2018-01-09
Feature/illuminate view

mosen:
Add support for Laravel Blade templates to KISSMVC and convert layout to Blade.

Attaches Illuminate/View via the Philo/Blade package from composer. Calling `$this->view()` from a controller will do the following:

- Setup the blade instance
- Assign variables that should be available to every view.
- Return the view so that your controller can assign more variables.

The partials head/foot have been wrapped into the layout called app.layout.
The navigation was extracted into navbar.
The stylesheet and javascript stacks allow a controller to inject css/js without changing custom_js or custom_css.

Also moved minimum PHP version to 7.0, and upgraded Laravel at the same time to 5.5

M


bochoven:
Not sure what happens, but it does not seem to work here..

- Login page comes up, but shows errors: 
```
Notice: Undefined index: role in /Users/Shared/munkireport-php/system/kissmvc.php on line 84
Notice: Undefined index: user in /Users/Shared/munkireport-php/system/kissmvc.php on line 89
```

- dashboard loads, but is missing header and footer (is there a blade template?):
```
<!-- Could not open /Users/Shared/munkireport-php//app/views/partials/head.php-->
```

bochoven:
Closing as we're moving to yaml views

PR #949, opened 2018-01-30
Fix installer postflight

MagerValp:
Bad quoting of $3 in the installer postflight caused 0600 permissions on the MunkiReport.plist preference file.

### DIFF
--- a/app/views/install/install_script.php
+++ b/app/views/install/install_script.php
@@ -62,7 +62,7 @@ EOF
 
 # Set munkireport preference
 function setpref {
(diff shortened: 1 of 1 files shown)


poundbangbash:
Just another data point.
Prior to running my `/Library/Preferences/MunkiReport.plist` was 
```
-rw-------  1 root  wheel  2982 Jan 20 23:51 /Library/Preferences/MunkiReport.plist
```

After running the bash install script it's now
```
-rw-r--r--  1 root  wheel  2982 Jan 30 17:09 /Library/Preferences/MunkiReport.plist
```

I also ran it thru creating a pkg. The postinstall script in the pkg looks good and after an install it has the correct permissions on the MunkiReport.plist.

-Eric

PR #953, opened 2018-02-02
Extension module TeamID breakout

poundbangbash:
With the advent of UAKEL and needing Team Identifiers and/or bundle ids of KEXTs there is going to be a need to easily access those values.  To help with that I took the existing extensions module and broke out the TeamID from the codesign value to make it easier to export and extract unique TeamIDs.

### Changes
* Renamed "codesign" column to "developer" in the database
* Added new column "teamid"
* Adjust UI to display Developer rather than Codesign ID
* Added TeamID widget in the report

I am waffling as to whether to keep the teamid in the Developer Name as well but I figured the duplication of information isn't that helpful when the listings shows the complete data.
Here's an example of UI showing both the old style and new style of data:

screenshot 2018-02-02 00 41 18

screenshot 2018-02-02 00 47 05

screenshot 2018-02-02 00 49 56


jbaker10:
This looks great! 👍

bochoven:
Hey @poundbangbash could you rename the migration to something like `2018_02_07_141313_extensions.php`?
I want to make sure it gets executed in the right order. 

BTW, there's a command line tool in munkireport for creating migrations:
```
build/addmigration app/modules/extensions
```

poundbangbash:
Changed the migration file name.

-Eric

PR #954, opened 2018-02-07
Experimental 10.13.4+ compatible module for MDM status

poundbangbash:
Yesterday 10.13.4 beta 2 introduced changes to the output `profiles` presents when checking MDM enrollment to better reflect the state of MDM enrollment. Of course this may change before 10.13.4 gets released but for now `profiles stats -type enrollment` now presents the following values:
```
Enrolled via DEP: No
Enrolled via DEP: Yes

MDM enrollment: No
MDM enrollment: Yes
MDM enrollment: Yes (User Approved)
```

This module captures that output for presenting in MR.
Since this data is only attainable from 10.13.4+ I'm up for suggestions/changes on how best to handle the module on OS versions < 10.13.4

I also think it would be good information to have on the client summary page but I wasn't able to decipher how to add the data from the computer record to that view.  Assistance there is appreciated as well.

-Eric


app/modules/mdm_status/mdm_status_model.php
```diff
@@ -0,0 +1,86 @@
...
+     * Process data sent by postflight
+     *
+     * @param string data
+     *
+     **/
+    public function process($data)
+    {
+    	if (strpos($data, '<?xml') === false) {
+    		// old style txt file data has been passed - throw an error.
+    		throw new Exception("Error Processing Request: old format data found, please update the security module", 1);
```
bochoven: no need for this

poundbangbash: Thanks, obviously leftovers from borrowed code.  I've removed it and updated the repo.

bochoven:
MDM status module moved: https://github.com/munkireport/mdm_status

PR #955, opened 2018-02-07
Changed power columns to be nullable in migration

tuxudo:
If columns are not nullable the migration may fail.


bochoven:
Please prepend month/day with 0 in migration name

PR #956, opened 2018-02-07
Fix bad datatype in usage_stats migration

tuxudo:
Datatypes need to be double or bigint, float is not able to properly store data.


bochoven:
Please prepend month/day in migration name with 0

app/modules/usage_stats/migrations/2018_02_08_000001_usage_stats.php
```diff
@@ -0,0 +1,260 @@
...
+    
+    private $tableName = 'usage_stats';
+    private $tableNamebackup = 'usage_stats_backup';
+    
+    public function up()
+    {
+        $capsule = new Capsule();
+        $migrateData = false;
+        
+        if ($capsule::schema()->hasTable($this->tableName)) {
```
bochoven: This will always be true if the previous migration has run?

bochoven: migratedata should check if a column has the correct type?

tuxudo: It'd be false if the module is initializing the table.

app/modules/usage_stats/migrations/2018_02_08_000001_usage_stats20180208000001.php
```diff
@@ -0,0 +1,260 @@
...
+    
+    private $tableName = 'usage_stats';
+    private $tableNamebackup = 'usage_stats_backup';
+    
+    public function up()
+    {
+        $capsule = new Capsule();
+        $migrateData = false;
+        
+        if ($capsule::schema()->hasTable($this->tableName)) {
```
bochoven: As this is the second migration, this will always be true, so the if statement can safely be removed (unless you want to cater for people deleting the table)

bochoven:
The migration is giving errors.

PR #957, opened 2018-02-07
Fixed WiFi nullable migration

tuxudo:


bochoven:
Please prepend month/day with 0

PR #960, opened 2018-02-09
User Sessions filtering 

tuxudo:
Added filtering options to user sessions. Can now filter based on events and if to keep historical data. Thanks to @poundbangbash for the ideas :D


bochoven:
It looks like some things got mashed together?

tuxudo:
Ugh, yep. I broke it. Let me fix it up

tuxudo:
Broke it :P

PR #963, opened 2018-02-09
Adjust database log view to use theme for readability.

poundbangbash:
The background color on the database log view is hard coded and makes it hard to read the content on some themes due to font/background contrast.  Removing the background color defaults back to the theme's CSS. I didn't see any issues in cycling thru the different themes.


poundbangbash:
Looks like this was addressed outside of this PR. Closing.

poundbangbash:
Wrong PR was closed.

PR #970, opened 2018-02-14
Added ssh status widget to report

poundbangbash:
Added SSH state stats widget to the Security report.


mosen:
Thanks, looks fairly straightforward. Did a quick checkout and test.

Issue #877, opened 2017-10-11
custom.js code runs at the wrong time


bartreardon:
as per request in [#munkiadmin](https://macadmins.slack.com/archives/C066NK3K5/p1507728590000450)

adding the following code to custom.js
```javascript
var user=$('nav i.fa-user')[0].nextSibling.nodeValue.trim();
var remoteip=$('span.mr-remote_ip').text();
$("a[href^='ssh://']").attr('href', 'ssh://'+user+'@'+remoteip);
```

appears to not run at the correct time

PR #976, opened 2018-02-18
Add option to use current username for ssh link

poundbangbash:
Using the method to get the current username from https://github.com/munkireport/munkireport-php/issues/877 this change allows an admin to use a custom ssh_link config to replace the username with the current user logged into Munkireport.  This is beneficial when directory based logins are used.

Adding `$conf['ssh_link'] = "ssh://%u@%s";` to config.php will result in the username being substituted in.  If the custom config is not in config.php the %u substitution does not take place.

-Eric

### DIFF
--- a/config_default.php
+++ b/config_default.php
@@ -230,6 +230,10 @@
 	| want the links, set either to an empty string, eg:
 	| $conf['vnc_link'] = "";
 	|
+	| If you want to authenticate with SSH using the currently logged in user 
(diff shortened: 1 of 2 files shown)


tuxudo:
Should this be mentioned in config_default.php?

poundbangbash:
Good idea.  Submitted.

childrss:
Awesome!!!

PR #979, opened 2018-02-21
Improved network module

tuxudo:
Things like DNS, VLANs, MTU, network media. Redid client tab with REST API
(Wi-Fi heading is clickable! :D)


bochoven:
I would like to add search domain too

tuxudo:
Closing PR, moved to module repo

PR #982, opened 2018-02-21
Fix DeployStudio migration

tuxudo:
All columns are a nullable string


bochoven:
I thought it should be a boolean?

tuxudo:
The data in the v2 database is a string and contains "YES" or "NO". The model and views would have to be changed if it was converted to a boolean

PR #985, opened 2018-02-22
Fixed DeployStudio processing on MR3

tuxudo:
Was throwing an error and not processing correctly.


bochoven:
I'm not using this module, so I trust you test these PR's 😄

PR #989, opened 2018-02-22
Add Events history time setting 

rfitzwater:
Adds ability to set time for how far back events are displayed from.


bochoven:
The events module moved here: https://github.com/munkireport/event

Issue #984, opened 2018-02-21
[FR] Add support to only allow management console access from certain IP's and/or CIDR ranges


danner26:
I've already built the code for this feature request, but I wanted to submit a issue for it to keep track. I will be forking the repo soon and adding my code (and updating it as I go along) so I can make a pull request when completed.

danner26:
Closing since merged to WIP

PR #990, opened 2018-02-22
Added support for server-side IP filtering

danner26:
This is in reference to Issue #984 

Basically this allows for you to disallow traffic to the management console from everywhere except certain IP addresses or CIDR ranges, while still allowing for clients to report to MunkiReport. 

When someone tries to access the MR URL, it will check the IP address of the client against the $conf['manage_ipwhitelist'] array of IP's or CIDR addresses.
If the client is whitelisted, it will let them through. Otherwise it will send them to the 403 page.

If the custom 403 page is setup with $conf['403_unauth'] then it will send them to that page. Otherwise it will send them to the webhost with /403.html at the end (the default location for a 403 page.


bochoven:
I like the idea, but this should be part of the auth mechanism

danner26:
That makes sense, I am assuming you mean implement it into the auth view instead of adding it to the head partial file correct?

bochoven:
Yes, same as the Captcha code. Maybe there's room for pre-auth plugins?

tuxudo:
It'd be cool to have two factor authentication. (But that's beyond the scope of this issue.)

danner26:
I agree, that would be awesome to have down the road.

danner26:
@bochoven I moved the check to the pre-auth section

app/controllers/auth.php
```diff
@@ -70,6 +70,10 @@ public function login($return = '')
             }
         }
 
+       if(!empty(conf('manage_ipwhitelist'))) { 
+         check_ip($_SERVER['REMOTE_ADDR']);
```
bochoven: You need to use `getRemoteAddress()` instead of `$_SERVER['REMOTE_ADDR']` to fix requests that come from a loadbalancer/proxy

config_default.php
```diff
@@ -774,6 +774,24 @@
...
+        | Whitelisting of IP addresses that can access the management interface 
+        |    (anything except for index.php?/report/ which is always allowed)
+        |  - You can provide either individual IP addresses (which will have /32 appended automatically)
+        |      or you can provide CIDR notation. See https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing for reference
+        |  - You can also provide a custom 403 page for traffic that does not have access to the management interface
+        |      Default: Value of $conf['webhost'] with 403.html added to the end
+        |
+        */
+        //$conf['manage_ipwhitelist'] = array('xxx.xxx.xxx.xxx', 'xxx.xxx.xxx.xxx/yy');
+        //$conf['403_unauth'] = 'http://fqdn/403.html';
```
bochoven: You can leave out the option for the 403 and use the built in error page?

app/views/partials/head.php
```diff
@@ -47,15 +47,13 @@
 		foreach($scripts as $script): ?>
 	<script src="<?php echo conf('subdirectory'); ?>assets/js/<?php echo $script; ?>" type="text/javascript"></script>
 <?php endforeach; ?>
-
```
bochoven: Please revert the whitespace changes

public/index.php
```diff
@@ -120,6 +120,7 @@ function fatal($msg)
 //===============================================
 require( SYS_PATH.'kissmvc.php' );
 require( APP_PATH.'helpers/site_helper'.EXT );
+require( APP_PATH.'helpers/ip_blocker.php' );
```
bochoven: I don't want the ip_blocker loaded for everyone, please load dynamically via autoloader in the auth controller.

app/helpers/ip_blocker.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
bochoven: This should be converted to a class in the `lib/munkireport` so that it can be autoloader on request.

config_default.php
danner26: My organiztion would like a custom page, but how about instead of sending it to the default 403 page for the server, send it to the the built-in one? This way, we can have a custom 403 page location, and everyone else that doesnt want one will be sent to the built in error page

danner26:
I will update everything shortly, please let me know if my comment on config_default.php is acceptable and if so I will update the ip_blocker.php and the config_default.php to reflect that change

config_default.php
bochoven: Ok, how about this:
```php
$conf['auth']['network'] = [
    'whitelist_ipv4' => [
        'xxx.xxx.xxx.xxx',
        'xxx.xxx.xxx.xxx',
    ],
    'redirect_unauthorized' => 'http://fqdn/403.html',
]
```
Where `redirect_unauthorized` is optional and defaults to the built-in 403 page

danner26:
Alright I integrated it with AuthHandler and the auth controller so it will do everything asked

I think I covered everything, but let me know if anything else needs changes!

danner26:
Is everything fixed or do you still need more changes?

bochoven:
Sorry about that, I did look at your changes. Things are better, but I would like to have the checking code in it’s own class and file like the other auth mechanisms.

danner26:
Sure! I'll work on moving that over and let you know. I don't want to upgrade our production instance until this is working fully.

danner26:
Please check the most recent push - I moved it completely out of the AuthHandler to its own class as requested

PR #999, opened 2018-03-04
Changed Time Machine and security modules to use FountationPlist

tuxudo:
instead of "plutil -convert xml1"


bochoven:
Did you test this? Especially the writing/uploading part?

tuxudo:
Yes, it all is working properly. It outputs the same plist and the model processes it properly.

PR #1001, opened 2018-03-05
Added more device types

tuxudo:


bochoven:
`keykoard` really? 😄

tuxudo:
Yea, I have some really awful 3rd party keyboards that are "USB Keykoard".

PR #1026, opened 2018-03-20
Fixed 32-bit apps widget name

tuxudo:
Missed this file name change :/


gmarnin:
Tested and confimed this widget works. Thanks @tuxudo

PR #1028, opened 2018-03-21
Added support for different ship to in GSX

tuxudo:
Bug found by @joncrain


joncrain:
Made the changes and verified that the GSX module now returns the correct results.

Issue #860, opened 2017-08-23
Profile Report sorted by installation count


ygini:
Profile Report present 3 informations (name, payloads and installation count) in two columns (name and payloads).

It could be nice to present the installation count in a distinct column to allow a sort by this value (like show me profiles installed only on few devices, to detect something unusual).

poundbangbash:
In an attempt to address this PR #1030 adds a Profile widget and report to group profiles by count. Clicking on the profile name links to the profile listing, filtering against the profile name.

Listings show raw data from the database and since there's not a database entry for the number of times  a profile has been reported in I thought this would be next best in offering a way to isolate the profiles based on install count.

-Eric

PR #1030, opened 2018-03-22
Profile count report

poundbangbash:
Attempt at #860

This creates a Profiles report and widget that lists the unique profiles an a count for each. 
Clicking on the profile name in the widget links to a filtered search in the listing against the profile name.

-Eric


gmarnin:
Tested on a local test environment. Looks good, nice addition.

PR #1032, opened 2018-03-22
phpdotenv support

mosen:
DO NOT MERGE, TEST ONLY

Needed to reorder index.php in order to load phpdotenv first via autoload.
Add example .env
Add requirement phpdotenv
Change config_default so that most defaults can be sourced from the environment.


bochoven:
Is this mergeable yet?

PR #1036, opened 2018-03-23
FileVault Escrow Issues#1035

danner26:
Issue #1036 with FileVault Escrow migration.  

When the table doesn't exist, you get the "base table or view not found" error. Simple clause to check for the table. Going to work on making it create the table if it is not currently present.


danner26:
Sorry about not cleanly pushing to this branch.. The top two commits f3215bc and 1a735c3 can be ignored.

bochoven:
The first migration should create the table. You should not have to recreate the table in this migration.

PR #1062, opened 2018-04-13
Realign the Login Page Box

gmarnin:
The text that pops over the lock icon at the login page is cutoff because the login box is too high up on the page. This PR pushes the login box down a bit so the text is visible. 

There maybe a more appropriate fix than a bunch of `<br>` tags but this works and looks good on a 27" iMac, 15" MBP and iPhone X.

Before: 
lock message cut off

After:
lock message readable

### DIFF
--- a/app/views/auth/login.php
+++ b/app/views/auth/login.php
@@ -1,5 +1,12 @@
 <?php $this->view('partials/head'); ?>
-
+<br>
+<br>
+<br>
+<br>
+<br>
+<br>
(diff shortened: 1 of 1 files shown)


bochoven:
Does it look ok on mobile?

gmarnin:
Only have an iPhone X and it looks good on it.

gmarnin:
Found an iPad 9.7" and it looks good.

PR #1063, opened 2018-04-13
Make the SIP check more resilient

barn-stripe:
The output of csrutil can be more than just one line of enabled or
disabled. If it's in a custom config, it will have '"enabled (Custom
Configuration)' in it, and many of the functions could still turned off.

Searching for just 'enabled' would make SIP seem active if just any of
the sub parts of SIP were enabled.


bochoven:
Please base your change on the wip branch and send your pull request against wip

barn-stripe:
done and done

PR #1068, opened 2018-04-16
Actually fix security.py module

barn-stripe:
Further to 13900e84, actually fix SIP reporting.

### DIFF
--- a/app/modules/security/scripts/security.py
+++ b/app/modules/security/scripts/security.py
@@ -29,15 +29,17 @@ def sip_check():
     """ SIP checks. We need to be running 10.11 or newer."""
 
     if float(os.uname()[2][0:2]) >= 15:
-        sp = subprocess.Popen(['csrutil', 'status'], stdout=subprocess.PIPE)
+        sp = subprocess.Popen(['csrutil', 'status'],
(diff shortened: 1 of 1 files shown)


barn-stripe:
I should probably write some tests, but I'm not sure how that would interact with the PHP ones.

cut and pasting that function and running it actually works.

```
import os
import subprocess

def sip_check():
 """ SIP checks. We need to be running 10.11 or newer."""

 badstring = 'System Integrity Protection status: disabled'

 if float(os.uname()[2][0:2]) >= 15:
     # sp = subprocess.Popen(['csrutil', 'status'],
     sp = subprocess.Popen(['echo', badstring],
                           stdout=subprocess.PIPE,
                           universal_newlines=True)
     out, err = sp.communicate()

     # just read the first line of the output, the
     # System Integrity Protection status: ....
     # search for a full stop, as custom configurations don't have
     # that there.
     first_line = out.split("\n")[0]
     if "enabled." in first_line:
         return "Active"
     else:
         return "Disabled"
 else:
     return "Not Supported"



x = sip_check()

print(x)
```

outputs

```
% python3 sip.py
Disabled
```

And removing the echo badstring

```
% python3 sip.py
Active
```

PR #1070, opened 2018-04-25
Rewrote displays module

tuxudo:
New script collects more data about displays and outputs to plist instead of txt


bochoven:
Big PR, not sure of all the implications. I saw some nested try blocks in the python that did not make sense to me. I also wonder if all the information is relevant as a lot of columns are stored but not shown

PR #1072, opened 2018-05-03
New Module: SentinelOne

poundbangbash:
This module uses the SentinelOne summary data from `sentinelctl summary` to keep track of:
* active-threats-present (boolean)
* agent-id (string)
* agent-running (boolean)
* agent-version (string)
* enforcing-security (boolean)
* last-seen (string)
* mgmt-url (string)
* self-protection-enabled (boolean)

The report shows the critical information in button form as well as the agent versions and management URL in case there are multiple sites.
screenshot 2018-05-02 23 13 26

The listing has colored values for easy overview of issues:
screenshot 2018-05-02 23 14 44

The client tab shows the data as well:
screenshot 2018-05-02 23 16 20

-Eric


bochoven:
The easiest way is to store the list of dicts as JSON in the database. You could store the count of files separately if you want to have a quick overview of quarantined files.
But if you want to create stats of quarantined files over the entire fleet, you would need to store them in a different database table.

poundbangbash:
Thanks for your insight. I had thought about creating a separate table but thought that was taking the easy, expensive way out. I'll work on getting that data ingested separately and presenting it accordingly.  For now the module is useful as is and will leave this here.

PR #1077, opened 2018-05-11
Added basic crypt checkin url

WardsParadox:
Added a checkin url to spoof a crypt server. This will respond with a 200 to stop the crypt client from attempting to escrow every 2 minutes.


WardsParadox:
Should be accessible at `munkireporturl/index.php?/module/filevault_escrow/checkin`
one would use ``munkireporturl/index.php?/module/filevault_escrow/` or `munkireporturl/index.php?/module/filevault_escrow` as the serverurl for crypt

bochoven:
May want to update the README as well

WardsParadox:
🤦‍♂️ I'll get on that.

PR #1087, opened 2018-05-26
Hotfix/migration patches

mosen:
Rewrote most migrations from v2.13 as v3 migrations.
Added LegacyMigrationSupport trait to support querying the old `migration` table, and setting new versions.
TODO: Testing and v2.15 migrations


mosen:
Tested against 2.12 and with new DB. Please merge into WIP.

Issue #815, opened 2017-05-22
FR: Configurable inactive timeframe for Client Activity widget


kevinmcox:
I think it would be helpful to have an easily configurable option to set the timeframe that is used to determine inactive clients in the Client Activity widget.

Admins could set the number of days that must past before a client is considered Inactive.

PR #1110, opened 2018-07-17
Support for custom client inactivity

tuxudo:
Added support for changing when a client goes inactive in the Client Activity widget. Configurable in config.php, default is 30 days. Resolves issue #815, requested about 3 times at PSU 2018

### DIFF
--- a/public/assets/locales/en.json
+++ b/public/assets/locales/en.json
@@ -106,6 +106,7 @@
         "comment": "Comments",
         "hour": "Clients per hour",
         "panel_title": "Client activity in the last month",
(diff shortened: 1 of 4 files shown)


public/assets/locales/en.json
```diff
@@ -106,6 +106,7 @@
         "comment": "Comments",
         "hour": "Clients per hour",
         "panel_title": "Client activity in the last month",
```
bochoven: Why not change panel_title instead of adding panel_title_custom?

tuxudo:
Requested changes made

PR #1115, opened 2018-07-27
Clarified the os dashboard widget description

gmarnin:


kevinmcox:
I just noticed this and to me it looked like a mistake that needed to be corrected.

Since there is no widget named "os breakdown" if an admin tries to use that in their configuration it won't work. They need to continue using "os" as always.

Personal opinion follows:

All other widgets listed in config_default.php are the actual text that an admin would use when configuring MunkiReport so it seems like this one should be no different.

If the change is really needed it seems like the widget itself should be renamed to **os_breakdown** instead of having the example config list extraneous text for this one widget but no others.

bochoven:
Ok with me, file a PR! :-D

kevinmcox:
OK, I'll include that when I reattempt my other PR after lunch.

PR #1120, opened 2018-07-31
macOSLAPS module CSS and config_default

tuxudo:
Module contains all needed scripts, securely stores password, verifies password is stored on server then deletes local cached plaintext copy, uses Joshua Miller's macOSLAPS script.


tuxudo:
To those that are testing this, you'll have to either drop and remake your database or manually add a column for the audit data.

bochoven:
This is a large PR, hopefully some people can test it and report back.

childrss:
Working on testing, adding some potential code... THANK YOU!!!

bochoven:
Not sure what you are trying to do, but things went south here I think

tuxudo:
Was trying to fix the conflict with config_default.php. Not exactly sure how it did what it did. I'll add it to the list of reasons why I don't like git.

tuxudo:
Removed module and changed PR to add required CSS and config_default.php options

PR #1121, opened 2018-08-01
Fixed menu sorting case sensitivity

tuxudo:


gmarnin:
I tested this and it works after I emptied the Safari cache

PR #1127, opened 2018-08-06
Update the usb module to collect device product ids

AaronBurchfield:
This change allows munkireport to collect the product ids from USB devices. An additional column is added to the usb table called `product_id`. Product ID will be available on the USB listing and on the USB Devices tab for each client. Please shout at me if there are any other changes I need to include.

### DIFF
--- a/app/modules/usb/migrations/2018_07_30_191408_usb_add_product_id.php
+++ b/app/modules/usb/migrations/2018_07_30_191408_usb_add_product_id.php
@@ -1,0 +1,29 @@
+<?php
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Capsule\Manager as Capsule;
+
+class UsbAddProductId extends Migration
+{
+    private $tableName = 'usb';
+
+    public function up()
+    {
+        $capsule = new Capsule();
+        $capsule::schema()->table($this->tableName, function (Blueprint $table) {
+          $table->string('product_id')->nullable();
+
+          $table->index('product_id', 'product_id_index');
+        });
+    }
+
+    public function down()
+    {
+        $capsule = new Capsule();
+        $capsule::schema()->table($this->tableName, function (Blueprint $table) {
+          $table->dropColumn('product_id');
+          
+          $table->dropIndex('product_id_index');
+        });
+    }
+}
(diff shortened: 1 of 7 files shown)


app/modules/usb/scripts/usb.py
```diff
@@ -40,6 +40,8 @@ def flatten_usb_info(array):
                 device['name'] = obj[item]
             elif item == 'vendor_id' or item == 'b_vendor_id':
                 device['vendor_id'] = obj[item]
+            elif item == 'product_id':
```
tuxudo: You'll have to make this line be `elif item == 'product_id' or item == 'a_product_id':` to prevent this module from breaking on older systems.

app/modules/usb/views/usb_listing.php
```diff
@@ -25,6 +25,7 @@
 			<th data-i18n="usb.device_speed" data-colname='usb.device_speed'></th>
 			<th data-i18n="usb.internal" data-colname='usb.internal'></th>
 			<th data-i18n="usb.media" data-colname='usb.media'></th>
+			<th data-i18n="usb.product_id" data-colname='usb.product_id'></th>
```
tuxudo: Don't forget to increase the `listing.loading` on line 34 when you add a column to the listing. The count should be `10`

AaronBurchfield:
@tuxudo thanks for the review! I've add the suggested changes.

app/modules/usb/migrations/2018_07_30_191408_usb_add_product_id.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    private $tableName = 'usb';
+
+    public function up()
+    {
+        $capsule = new Capsule();
+        $capsule::schema()->table($this->tableName, function (Blueprint $table) {
+          $table->string('product_id')->nullable();
+
+          $table->index('product_id', 'product_id_index');
```
bochoven: Please prepend the index name with 'usb_' or don't specify the index name and let eloquent generate the name.

AaronBurchfield: I've removed the name of the index, thanks!

bochoven:
The USB module has moved here: https://github.com/munkireport/usb

PR #1128, opened 2018-08-09
Added version of database server

WardsParadox:
Please test with mysql, I wasn't able to do that yet.


gmarnin:
I tested this PR against my production mysql db. The version is correctly reported on the `Munkireport System Status` page. Looks good. Thanks @WardsParadox

WardsParadox:
Need translation help with the new Version text as well.

Issue #1008, opened 2018-03-11
Display Machine Group name


ofirgalcon:
It would be useful to see the computer group name wherever a computer is listed - including the computer details page. Currently the group is only displayed there, but as a numerical id, which is less helpful.

bochoven:
Should not be too hard to implement

ofirgalcon:
Would be nice to have it shown in lists as well.

MichalMMac:
It would be great to be able to see machine group in computer summary.

PR #1129, opened 2018-08-09
Show machine group name instead of number

tuxudo:
Resolves issue #1008


ofirgalcon:
Would be great to see the groups in the various computer listings.

PR #1131, opened 2018-08-10
Add support for default theme

tuxudo:
Config option to set the default theme for new users/logins. May need to clear cache and local data if already logged in.


tuxudo:
Requested changes made

PR #1109, opened 2018-07-12
Add Jamf module

tuxudo:
Adds support for pulling data from a Jamf/JSS server

PR #1134, opened 2018-08-15
Support for modular admin views

tuxudo:
PR split from PR #1109


bochoven:
I like the idea, but not the implementation :-D. It would be nice if the menu items can be fetched as json and then put in the menu by calling `addMenuItem`.

tuxudo:
This feature was added with the release of 5.3.2

PR #1143, opened 2018-08-29
Caching module config_default.php update

tuxudo:
Added proper support for 10.13.4+


tuxudo:
Module removed to its own repo, PR changed for config_default.php changes

Issue #1147, opened 2018-09-10
Feature request - Filters key shortcut


ofirgalcon:
I like using D and C shortcuts. How about F for the Machine Group Filter?

tuxudo:
Give PR 1150 a test, f to show the filter popup and shift-f to hide it again

ofirgalcon:
I still have no clue about git and how to merge code from various pathces, test modules.

ofirgalcon:
Has this feature been merged into v4? It doesn't seem to work.

bochoven:
It's merged now. Available in wip

PR #1150, opened 2018-09-11
Added F for machine filter popup

tuxudo:
Resolves feature request #1147


bochoven:
I like it, but please remove the code to close the filter modal. esc already does that.

PR #1179, opened 2018-11-09
Add laps options to config_default

tuxudo:
and cleans up formatting of config_default


bochoven:
I think this should be moved to the LAPS module

PR #1199, opened 2018-12-03
Removing disk_report from default dashboard

### LINKED ISSUE #1198: Remove disk_report from default dashboard
I believe `disk_report` used to be a standard included module, but that no longer seems to be the case. However it is still included in the default dashboard which results in a `disk_report This widget cannot be found` warning.

joncrain:
Should resolve #1198


bochoven:
config_default will be gone soon and I think it's better to add disk_report as a default module

PR #1228, opened 2019-02-21
Make showing help link default to true

WardsParadox:
Not sure if all agree, but I think having the (?) symbol be visible by default makes sense.


kevinmcox:
I think it is a worthwhile change that could help new MR users while being easily disabled by experienced users who don't want to see it.

PR #1235, opened 2019-03-12
Change favicon to colored blocks

tuxudo:
Changed favicon to use a colored blocks style based off of the colored block widget on the dashboard


tuxudo:
Colored Blocks

WardsParadox:
While I do love (and hate think tombstone 😆 ) the logo I made, I think for a favicon, bookmark icon, etc it's not as clear at smaller sizes. Has my vote!

bochoven:
Why the numbers?

WardsParadox:
I think the numbers represent the data that you get on the dashboards.

tuxudo:
@WardsParadox is correct. Just numbers showing what the colored blocks can look like

bochoven:
If we're going to change the icon, I would prefer something else! 😄

PR #1242, opened 2019-03-26
Add x509 cert and private key to SP array for SAML

cwhits:
following the onelogin/php-saml docs (https://github.com/onelogin/php-saml/blob/master/settings_example.php#L66-L67) if you want to have your cert outside of the vendor/onelogin/php-saml/certs folder you should use x509cert and privateKey in the SP array. This PR moves them there and keeps their defaults blank. 

I'm not totally sure that `'cert_directory' => env('AUTH_SAML_CERT_DIR', local_conf('certs/')),` is working correctly as I can't have any value in there not use the local certs dir in the onelogin/php-saml directory.

### DIFF
--- a/app/config/auth/saml.php
+++ b/app/config/auth/saml.php
@@ -4,6 +4,8 @@ return [
   'sp' => [
     'NameIDFormat' => env('AUTH_SAML_SP_NAME_ID_FORMAT', 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress'),
     'entityId' => env('AUTH_SAML_SP_ENTITY_ID', ''),
+    'x509cert' => env('AUTH_SAML_SP_X509_CERT', ''),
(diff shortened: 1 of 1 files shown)


cwhits:
I guess I should also add that a good reason to have them outside of vendor/onelogin/php-saml is that directory gets blown away every update.

cwhits:
uh, I noticed that this is already in master but it's not in WIP. did this PR get made against the wrong branch?

https://github.com/munkireport/munkireport-php/pull/1214

bochoven:
I probably f*d up that merge

bochoven:
That said, there is already support for storing the x509 cert outside the vendor directory. I'm afraid it did not get documented.

https://github.com/munkireport/munkireport-php/tree/master/local/certs

cwhits:
@bochoven is that only for the IdP cert or the SP certs too?

PR #1252, opened 2019-06-12
Remove duplicate hosts key

homebysix:
There were two instances of the `hosts` key in the config_to_env.yml file. This change removes one of them.

PR #1253, opened 2019-06-12
Add pre-commit hooks for PHP linting, Python Black, and other basics

homebysix:
This change would implement a [Pre-Commit](https://pre-commit.com) configuration that prevents contributors from committing unless certain basic tests pass. As configured here, the tests include:

- [Black](https://black.readthedocs.io/en/stable/) formatting for Python files
- All PHP files pass `php -l` linting check
- No added files exceed 100KB
- No leftover merge conflicts within files
- UNIX style line endings only

Additional hooks have been included in the config file but remain commented out, because they will require a bit more work to initially implement. These future tests could include:

- Eliminating unnecessary trailing whitespace
- Standardizing file endings
- YAML syntax checking (which would catch issues like the one fixed in #1252)
- Verifying executable shebangs

Many other [pre-commit hooks](https://pre-commit.com/hooks.html) are available, some of which may also be useful and can be added later if the maintainers find them useful.

In order to use pre-commit, each contributor would need to [install pre-commit](https://pre-commit.com/#install), then run `pre-commit install` to activate the hooks. A similar system is now in use for the [AutoPkg project](https://github.com/autopkg/autopkg/blob/master/CONTRIBUTING.md#use-pre-commit-to-set-automatic-commit-requirements).

Thanks for considering!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
   - id: check-case-conflict
   # - id: check-docstring-first
   # - id: check-executables-have-shebangs
+  - id: check-json
   - id: check-merge-conflict
   # - id: check-yaml
   # - id: end-of-file-fixer
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,7 @@ repos:
   # - id: check-executables-have-shebangs
   - id: check-json
   - id: check-merge-conflict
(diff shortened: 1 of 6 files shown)


bochoven:
@homebysix would you be able to resolve the merge conflicts?

homebysix:
It was easier to create a new PR rather than resolve the conflicts in this one, so I created #1299.

PR #1269, opened 2019-08-03
fix one forgotten case of curl quoting in install script

doits:
only materialized itself when additional headers where set: in this case
they where not passed quoted correctly


bochoven:
Please send PR against thewip branch

doits:
@bochoven done

PR #1293, opened 2019-10-29
Helper script for managing MunkiReport installs

lcsees:
Recently created a tool (`build/mr_upgrade.py`) that we are utilizing in our environment for managing a MunkiReport install that we would like to share back with the general repository if it may be deemed of value. Its use is to upgrade a local version of MunkiReport to the latest version, performing backups, composer updates, and database migrations as necessary, with options to rollback to a previous version or restore from database backups.

If you have any suggestions for improvement, issues, or comments please feel free to let me know. Any requirements are detailed in `build/requirements.txt`. Thank you! :)


bochoven:
Can’t test, but it looks like a nice addition!

PR #1299, opened 2019-11-12
Add pre-commit hooks for PHP linting, Python Black, and other basics (conflict resolved)

homebysix:
(Replaces #1253.)

This change would implement a [Pre-Commit](https://pre-commit.com) configuration that prevents contributors from committing unless certain basic tests pass. As configured here, the tests include:

- [Black](https://black.readthedocs.io/en/stable/) formatting for Python files
- All PHP files pass `php -l` linting check
- No added files exceed 100KB
- No leftover merge conflicts within files
- UNIX style line endings only

Additional hooks have been included in the config file but remain commented out, because they will require a bit more work to initially implement. These future tests could include:

- Eliminating unnecessary trailing whitespace
- Standardizing file endings
- YAML syntax checking (which would catch issues like the one fixed in #1252)
- Verifying executable shebangs

Many other [pre-commit hooks](https://pre-commit.com/hooks.html) are available, some of which may also be useful and can be added later if the maintainers find them useful.

In order to use pre-commit, each contributor would need to [install pre-commit](https://pre-commit.com/#install), then run `pre-commit install` to activate the hooks. A similar system is now in use for the [AutoPkg project](https://github.com/autopkg/autopkg/blob/master/CONTRIBUTING.md#use-pre-commit-to-set-automatic-commit-requirements).

Thanks for considering!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
   - id: check-case-conflict
   # - id: check-docstring-first
   # - id: check-executables-have-shebangs
+  - id: check-json
   - id: check-merge-conflict
   # - id: check-yaml
   # - id: end-of-file-fixer
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,7 @@ repos:
   # - id: check-executables-have-shebangs
   - id: check-json
   - id: check-merge-conflict
+  - id: check-xml
   # - id: check-yaml
   # - id: end-of-file-fixer
   - id: mixed-line-ending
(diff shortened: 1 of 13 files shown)


rickheil:
I love this idea, thank you for filing this PR. Is the .pre-commit and config generic enough that module maintainers could easily clone those same standards? (asking selfishly)

gmarnin:
I worry that this sets the bar too high for contributions. I don't want to prevent folks from summiting PRs because that can't or won't jump this hurtle.

homebysix:
@gmarnin, sorry I never replied to your comment. The pre-commit hooks only run on local Git repos, not on GitHub.com (unlike CI/CD, for example). So while I think this PR justifiably raises the bar for multi-commit contributions that are worked on locally before being pushed, the PR process is unchanged for small contributions that can be done using GitHub.com's inline editor.

PR #1347, opened 2020-05-18
Local module directory

joncrain:
Adding a default search path for custom modules in the local directory.

### DIFF
--- a/app/config/app.php
+++ b/app/config/app.php
@@ -158,7 +158,7 @@ return [
     | Filesystem paths to search for modules
     |    replaces the implicit 'custom' module path
     */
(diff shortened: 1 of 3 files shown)


joncrain:
So close. I've noticed that if I configure an alternative local_directory_path, if I leave off the trailing slash, it does not work, though the other local config does.

app/config/app.php
```diff
@@ -158,7 +158,7 @@ function (){
     | Filesystem paths to search for modules
     |    replaces the implicit 'custom' module path
     */
-    'module_search_paths' => env('MODULE_SEARCH_PATHS', []),
+    'module_search_paths' => env('MODULE_SEARCH_PATHS', [env('LOCAL_DIRECTORY_PATH', APP_ROOT . 'local/') . 'modules']),
```
bochoven, line 161: Don't need this anymore

PR #1349, opened 2020-05-20
Add Module Marketplace

tuxudo:
An admin page that details installed modules and available widgets


joncrain:
My initial install did not pull in all database columns for some reason. After I reran my dev setup script, it started working.
image

Also the enabled modules does not seem to be working for me. When I have nothing in my .env for "MODULES", then all show up as enabled. After I added a few in the .env, all show up as no.

Also, the path for core modules has an added `/` `/munkireport_dev//vendor/munkireport/supported_os`

tuxudo:
The double slash is how MR returns the data, I'll correct it so it doesn't show. Not sure why your modules are acting that way. Will look into that more tomorrow.

joncrain:
If I surround my module list with `'`, it works, `"` does not work. 

Also, how are you pulling date downloaded? It looks to be the same as date updated.

tuxudo:
If your modules are up to date, then downloaded and uploaded will be the same. May change that to the timestamp of the model file itself. 

The `'` and `"` is in .env? It's just going off of what MR gives it

joncrain:
Yeah, I'm not sure what's going on. It hasn't been consistent for me. I'll check it again in the morning though.

tuxudo:
Fixed the double slash, first loading of the MM on a new install, fixed some deprecation warnings, fixed enabled logic

app/controllers/Module_marketplace.php
```diff
@@ -0,0 +1,425 @@
...
+use Symfony\Component\Yaml\Yaml;
+use munkireport\models\Cache;
+
+class Module_marketplace extends Controller
+{
+    private $moduleMarketplace;
+
+    public function __construct()
+    {
+        $this->moduleMarketplace = getMrModuleObj();
```
bochoven, line 16: // Check authorization
        $this->authorized() || jsonError('Authenticate first', 403);
        $this->authorized('global') || jsonError('You need to be admin', 403);

tuxudo: Added

tuxudo:
Not yet ready to merge, working on removing the need for the database part of it

tuxudo:
PR ready to be reviewed and merged.

bochoven:
If the database is not needed anymore, the migration is also not necessary?

tuxudo:
Database elements and migration removed

PR #1369, opened 2020-07-01
munki + munkireport postflight - update shebang

carlashley:
Updated the munkireport postlight script shebang to point to the one managed by munki.


bochoven:
Despite the name, MunkiReport is also used in environments where munki is not installed. So changing the shebang would impact those environments in a bad way.

PR #1412, opened 2021-05-06
Fixed #1411 - Fix Deprecation warning: Required parameter follows optional

### LINKED ISSUE #1411: Deprecated: Required parameter $allowedModules follows optional parameter $skipInactiveModules
app\lib\munkireport\Modules.php on line 381

mantis:
Fix Issue #1411 php warning by making optional parameter required. This function only has 1 callee, and optional parameter is always supplied


bochoven:
This is issue is already resolved in 5.x

PR #1413, opened 2021-05-09
add auth based on server env var

barteklk:
Adds auth mechanism `ENV` that is based on an environment variable set by a server

This allows MunkiReport's auth to be configured in alignment with standard auth mechanisms available in some HTTP servers, like e.g. Apache HTTPd.  With auth set to `ENV`, MunkiReport knows that it has to look for a value of an environment variable to determine the name of the user logged in.  The name of the environment variable is configurable, set to `REMOTE_USER` by default.

### DIFF
--- a/app/config/auth/env.php
+++ b/app/config/auth/env.php
@@ -1,0 +1,5 @@
+<?php
+
+return [
+    'env_user_var'            => 'REMOTE_USER',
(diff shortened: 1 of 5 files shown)


barteklk:
Resolves feature request #1415

PR #1482, opened 2022-09-02
Updated method to get CPU arch to get it from the full uname string

jc0b:
This PR changes how the main runner grabs the CPU arch, as using `os.uname()[4]` is unreliable when used with Rosetta 2 on Apple Silicon.

Instead, we just search the uname string provided by `os.uname()[3]`, and lowercase it to match existing data.

I have tested this with the arm64 and x86_64 strings, but I don't have a physical machine to test i386 with 😅


tuxudo:
Superseded by PR #1498

bochoven:
Already merged in 5.x

Issue #1515, opened 2023-06-23
http auth basic broken on 5.8.0 with python 3.10


jstaubr:
Hi,

after upgrading to the latest release with newer python I'm unable to upload new data to the server from my test client because I get 401 unauthorized. I'm using an MDM profile with MunkiReport payload containing the http basic token.

Downgrading back to 5.7.1 with python 2.7 makes the authentication and upload work.

**5.8.0 client**
```
    Requesting printer
    Authentication challenge for Host: xxxxx.cz Realm: None AuthMethod: NSURLAuthenticationMethodServerTrust
    Allowing OS to handle authentication request
    Authentication challenge for Host: xxxxx.cz Realm: Restricted Content AuthMethod: NSURLAuthenticationMethodHTTPBasic
    Allowing OS to handle authentication request
    Authentication challenge for Host: xxxxx.cz Realm: None AuthMethod: NSURLAuthenticationMethodServerTrust
    Allowing OS to handle authentication request
    Status: 401
    Headers: {'Server': 'nginx/1.20.2', 'Date': 'Fri, 23 Jun 2023 07:08:26 GMT', 'Content-Type': 'text/html; charset=iso-8859-1', 'Connection': 'keep-alive', 'Www-Authenticate': 'Basic realm="Restricted Content"', 'Content-Length': '468'}
    ERROR: https://xxxxx.cz/index.php?/report/hash_check failed, HTTP returncode 401 (unauthorized)
```
**Server log**
```
[23/Jun/2023:09:03:08 +0200] "POST /index.php?/report/hash_check HTTP/1.0" 401 703 "-" "Python/3.10.9 CFNetwork/1465.1 Darwin/23.0.0"
[23/Jun/2023:09:03:08 +0200] "POST /index.php?/report/hash_check HTTP/1.0" 401 703 "-" "Python/3.10.9 CFNetwork/1465.1 Darwin/23.0.0"
```

**5.7.1 client**
```
    Requesting printer
    connection_willSendRequestForAuthenticationChallenge_
    Authentication challenge for Host: munki-report.sbeta.cz Realm: None AuthMethod: NSURLAuthenticationMethodServerTrust
    Allowing OS to handle authentication request
    ...
    Sending items (109KB)
```
**Server log**
```
[23/Jun/2023:09:14:22 +0200] "POST /index.php?/report/hash_check HTTP/1.0" 200 401 "-" "Python/2.7.18 CFNetwork/1465.1 Darwin/23.0.0"
[...]

PR #1516, opened 2023-06-26
Update reportcommon.py

tuxudo:
Fix config bug discovered in issue #1515


bochoven:
Was this key always wrong? Or was there a reason it did not contain munki?

tuxudo:
I changed this key in changing from Python 2 to 3 to a new incorrect key. Documentation, existing configs, and another script all reference it as being `UseMunkiAdditionalHttpHeaders`

PR #1517, opened 2023-06-29
Check for new MunkiReport version on Module Marketplace

tuxudo:
Checks for new version of MunkiReport when visiting Module Marketplace. Uses GitHub API for version checking within browser page.


tuxudo:
Button links to GitHub release page for version show, only shows if text and button if there is a new version

image

PR #1518, opened 2023-06-29
Fix additional key

tuxudo:
Additional fix for issue #1515 and PR #1516


tuxudo:
This PR is ready for MR v5.8.1 @bochoven

PR #1526, opened 2023-08-03
Update github-registry.yml

sphen13:
This is to enable multi-arch image creating with github actions. I am not clear on whether this need to go to the main branch, both main and 5.x or just 5.x. please review. it is working well in my fork. Thanks!


sphen13:
goin to resubmit to 5.x branch as that is where it should be after testing

PR #1533, opened 2023-08-11
Set permissions on cache directory

tuxudo:
Sets permissions on cache directory to 600 `drw------- root wheel` to prevent non-admin users from accessing potentially sensitive cache files.


bochoven:
Not sure if this is a good idea, munkireport may not run as root

tuxudo:
Most (all?) the modules are written in such a way that assume it is run as root or a user with root privileges, due to the higher level of access needed by some of the commands used by them. `munkireport-runner` (in MR 5.8+) has a check in itself to check that it's run as root that prevents a non-admin from doing anything other than getting the current installed version of MunkiReport: https://github.com/munkireport/munkireport-php/blob/5.x/public/assets/client_installer/payload/usr/local/munkireport/munkireport-runner#L73. 

The cache directory has the current permissions of `drwxr-xr-x root wheel` requiring that as currently written `munkireport-runner` must be run as root to write to that directory, but allowing anyone to read the files within.

bochoven:
Ok, makes sense!

PR #1546, opened 2024-08-09
CSP generator

ofirgalcon:
This change add a CSP header. It may need some looking over but it seems to work.


ofirgalcon:
This version removed the report_url

PR #1563, opened 2025-09-07
Update head.php

ofirgalcon:
Add font awesome v6


ofirgalcon:
doesnt work :(

ofirgalcon:
Fixed. If you use CSP header then add cdnjs.cloudflare.com to default-src and style-src

tuxudo:
Font Awesome 6 is a large breaking change. I do not want to maintain both FA v4 and v6 in the same codebase. This is not a thing that will be merged into MR v5, but could be revisited for v6.

ofirgalcon:
It doesnt break though, it's fully backwards ccompatible.

PR #1565, opened 2025-10-21
Add deny-if-empty feature to ENV authentication

overhacked:
When using `ENV` environment variable authentication, add option to deny authentication if the environment variable (default `REMOTE_USER`) is empty or missing from the environment. When a web server defaults to `REMOTE_USER` being empty for unauthenticated sessions, this prevents a user with "blank username" from being logged into MunkiReport. This also allows for fallback to another authentication method; for example:
1. Try `ENV` auth to check if `REMOTE_USER` is set for users accessing MunkiReport via Tailscale Serve, which can set an authenticated username request header.
2. Fall back to `LDAP` or `SAML` auth if users are accessing MR not via Tailscale.

*Notes*:
* Disabled by default, so the previous `ENV` behavior is preserved, and this is not a breaking change
* A present but blank environment variable returns `unauthorized`, and a missing environment variable returns `failed`; this would aid in debugging.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -79,6 +79,7 @@ SITENAME="MunkiReport"
 #
 # AUTH_METHODS can be one of
 # - "NOAUTH": No authentication
+# - "ENV": Environment variable (without password) Authentication
 # - "LOCAL" : Local Users defined as .yml in the "users" folder
 # - "LDAP": LDAP Authentication
 # - "AD": Active Directory Authentication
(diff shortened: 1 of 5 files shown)


overhacked:
I've already posted some [Wiki changes related to this pull request](https://github.com/munkireport/munkireport-php/wiki/_compare/1a2340d58b3979396402f87aa7424d16fabb7485...648768773b8551e16a8361f5de3a3a5d4f3aceb1). I apologize that the documentation got ahead of the PR; I was writing and forgot that saving to the wiki isn't subject to review. If this doesn't get merged, I'll be happy to go and revert.

PR #1572, opened 2026-04-28
Update vlucas/phpdotenv to version  ^5.0 in Composer.json

precursorca:
"vlucas/phpdotenv": "^5.0", 
Required to avoid depecration warnings with PHP 8.1+


precursorca:
NO, No, No.
Turns out that going from phpdotenv 4 to 5 prevents logins! 
So keep the version at 4.

